PR #4, opened 2016-08-21
Added DOI/ORCID validator

kyleniemeyer:
The `reference` entry of a ChemKED YAML file is now checked using a new `_validate_isvalid_reference` function in the `OurValidator` class (along with `isvalid_reference: true` added to the schema). 

If present, it uses the reference DOI and Crossref DOI to validate the publication metadata (journal, year, volume, pages) and author names (currently just last name).

In addition, if any author ORCIDs are present, it confirms those using the ORCID API.


bryanwweber:
The only concern I have here is what happens if the user doesn't have network access. We need to fail gracefully in that case.

chemked/chemked_schema.yaml
```diff
@@ -1,8 +1,9 @@
 # Common reference for authors' information
-author:
+author: &author
```
bryanwweber: The `author` field should be required. I also found that it didn't work properly when I put the reference here, but it passes the tests, so... who know!

chemked/validation.py
```diff
@@ -47,6 +49,49 @@
                   'time': 'second',
                   }
 
+def compare_name(given_name, family_name, question_name):
+    """Compares a name in question to a specified name separated into given and family.
+
+    The name in question ``question_name`` can be of varying format, including
+    "Kyle E. Niemeyer", "Kyle Niemeyer", "K. E. Niemeyer", "KE Niemeyer", and
+    "K Niemeyer". Other possibilities include names with hyphens such as
+    "Chih-Jen Sung", "C. J. Sung", "C-J Sung".
```
bryanwweber: `C-J Sung` isn't tested, only `C.-J. Sung` is tested. Are they the same?

chemked/validation.py
```diff
@@ -47,6 +49,49 @@
...
+    """
+    # lowercase everything
+    given_name = given_name.lower()
+    family_name = family_name.lower()
+    question_name = question_name.lower()
+
+    # split name in question by , <space> - . ! ? :
```
bryanwweber: Do people really have `!?:` in their names?

chemked/validation.py
```diff
@@ -47,6 +49,49 @@
...
+    name_split = list(filter(None, re.split("[, \-.!?:]+", question_name)))
+    first_name = [name_split[0]]
```
bryanwweber: What if people put in `Niemeyer, Kyle E` into the ChemKED file?

chemked/validation.py
```diff
@@ -174,26 +174,26 @@ def _validate_isvalid_reference(self, isvalid_reference, field, value):
...
-                    elif author_match.get('ORCID'):
-                        # still validate if present in file
-                        res = orcid_api.search_public('orcid:' +
-                                                      author_match['ORCID']
-                                                      )
-                        if res['orcid-search-results']['num-found'] == 0:
-                            self._error(field, 'ORCID incorrect for ' +
-                                        author_match['name']
-                                        )
+                # else:
```
bryanwweber: Any reason to comment this out instead of deleting?

chemked/tests/test_validation.py
```diff
@@ -119,6 +119,31 @@ def test_invalid_DOI(self):
...
+                   )
+        assert v.errors['file-author'] == 'ORCID incorrect or invalid for Kyle Niemeyer'
+
+    def test_invalid_ORCID_name(self):
+        """Test for proper response to incorrect name with ORCID.
+        """
+        v.validate({'file-author': {'ORCID': '0000-0003-4425-7097',
+                   'name': 'Bryan Weber'}}, update=True
+                   )
+        assert v.errors['file-author'] == 'name incorrect, should be Kyle Niemeyer'
```
bryanwweber: I think this error should be `Name and ORCID do not match. Name supplied: Bryan Weber. Name associated with ORCID: Kyle Niemeyer` or something similar. It could also be that the ORCID is wrong :smile:

chemked/chemked_schema.yaml
kyleniemeyer: Let's try your solution for `value-unit` with `author`, but wait until this PR is done (maybe do in your PR #7.

chemked/validation.py
kyleniemeyer: Added tests for this in fa8470a

chemked/validation.py
kyleniemeyer: Removed those unnecessary characters from the splitting with 7fd0c2c.

It does split on periods, so that `'K.E. Niemeyer'` becomes `['K', 'E', 'Niemeyer']`

chemked/validation.py
kyleniemeyer: Well now you're just being difficult 😿

Right now it assumes the names are written in the correct order. I could have it look for a comma, and consider anything with a comma written as above, as in `'Niemeyer, Kyle E'` -> `['Niemeyer', 'Kyle', 'E']` -> `['Kyle', 'E', 'Niemeyer']` and then handled normally. This seems pretty easy, so I'll do this.

Since the period is at the end of that string, it would just be removed.

*or* we could just require names formatted correctly 😀

kyleniemeyer: Alright, just added support and tests for this via 1c7b30b

chemked/validation.py
kyleniemeyer: Just to temporarily save it until sure I didn't need it 😁

kyleniemeyer: Before separating the ORCID validator off the DOI validator, this function checked if the ORCID in reference metadata matched the given ORCID. For older articles this won't happen much, but now with article submissions I think publishers are pushing people to get ORCIDs and associate them with papers (and rightfully so).

If an author has an ORCID associated with the reference, we can (a) quickly check that it matches the given ORCID, if one is given; or (b) suggest the ORCID be added if it is not given.

kyleniemeyer: OK, added that stuff with tests in 04cb2f9

chemked/tests/test_validation.py
kyleniemeyer: OK, improved this error message via b183c9b

PR #13, opened 2016-08-25
Add tests for incompatible units and quantities

bryanwweber:
Add some more tests. Make coverage great again!  Before this gets merged, I want to exclude setup.py from the coverage.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -11,3 +11,4 @@ exclude_lines =
 ignore_errors = True
 omit =
     chemked/tests/*
+    setup.py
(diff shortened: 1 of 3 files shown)


kyleniemeyer:
I'm somewhat disappointed that the commit message isn't `making coverage great again`.

kyleniemeyer:
Looks like we can use `codecov.yml` to exclude paths or files. Let me give that a shot

bryanwweber:
Only commits that get us to 100% coverage should be called `making coverage great again` 😛

PR #16, opened 2016-08-30
Get a DataFrame as output

bryanwweber:
Adds a method to `ChemKED` to get a Pandas DataFrame as output. Only imports pandas if this method is called, so pandas is not required to run the whole package. Adds tests that should skip if pandas is not found.

<s>Tests are currently failing because of a server issue at CrossRef: https://github.com/CrossRef/rest-api-doc/issues/147</s>


bryanwweber:
I also forgot I was supposed to look at `@staticmethods` etc. I think that deserves a different PR.

chemked/chemked.py
```diff
@@ -95,6 +95,102 @@ def validate_yaml(self, properties):
...
+                    * Apparatus:Kind
+                    * Apparatus:Facility
+                    * Apparatus:Institution
+                Only the first author is printed when ``Reference`` or ``Reference:Authors`` is
+                selected because the whole author list may be quite long.
+
+        Returns:
+            DataFrame: Contains the information regarding each point in the ``datapoints`` attribute
+        """
+        import pandas as pd
```
kyleniemeyer: Generally I prefer to have all imports at the top of the file, but I think an exception is warranted here.

Do you need to abbreviate as `pd`, though?

chemked/tests/test_chemked.py
```diff
@@ -39,6 +39,103 @@ def test_datapoints(self):
...
+
+        ChemKED(dict_input=properties)
+
+    def test_unallowed_input(self, capfd):
+        file_path = os.path.join('testfile_required.yaml')
+        filename = pkg_resources.resource_filename(__name__, file_path)
+        with open(filename, 'r') as f:
+            properties = yaml.safe_load(f)
+
+        properties['experiment-type'] = 'Ignition Delay'  # should be ignition delay
```
kyleniemeyer: Do we want to be case sensitive?

chemked/chemked.py
```diff
@@ -95,6 +95,102 @@ def validate_yaml(self, properties):
...
+                    * Reference:Pages
+                    * Apparatus:Kind
+                    * Apparatus:Facility
+                    * Apparatus:Institution
+                Only the first author is printed when ``Reference`` or ``Reference:Authors`` is
+                selected because the whole author list may be quite long.
+
+        Returns:
+            DataFrame: Contains the information regarding each point in the ``datapoints`` attribute
+        """
```
kyleniemeyer: maybe add example usage here? like 
```
Example:
    >>> df = ChemKED(yaml_filename).get_dataframe()
```

kyleniemeyer:
I made a few minor comments, but otherwise I think it looks good.

chemked/chemked.py
bryanwweber: Yes, me too, but since this is not essential to the functioning of the package as a whole, I don't think it should be a dependency. Is there a reason not to abbreviate?

chemked/tests/test_chemked.py
bryanwweber: This is how the validation is currently written. If we want to change it, I support that, and the validator needs to change. This is just testing that the value is not allowed and could be any invalid string.

chemked/chemked.py
kyleniemeyer: Just for clarity when reading—one more thing to keep in your head. `pd` isn't much shorter than `pandas` anyways. I'm similarly in favor of just using `numpy` instead of `np`.

chemked/tests/test_chemked.py
kyleniemeyer: I do think we should change the validator, but this is fine for now since it matches the current behavior.

chemked/chemked.py
bryanwweber: I meant to respond to this... I actually find it easier to read with the abbreviation, perhaps because they're familiar to me, but I think also because the abbreviations take up fewer characters to distract from what the code is doing. Anyways, you merged it, so I assume its OK to leave it.

PR #18, opened 2016-08-30
Create contributing doc

kyleniemeyer:
Also added some docs to `validation.py` functions, to set a good example.


bryanwweber:
I think the `CONTRIBUTING.md` file should go in a `.github` folder per https://github.com/blog/2111-issue-and-pull-request-templates. We should also create PR and Issue templates.

kyleniemeyer:
I'd prefer to leave `CONTRIBUTING.md` in the top level, so that it's obvious rather than a hidden file. I do agree with the issue and PR templates, though.

PR #25, opened 2016-11-19
Added better author validation and tests

### LINKED ISSUE #11: Author validation should check for extra authors from YAML
Right now, we only check if authors are missing in the YAML. We should also check whether extra authors have been provided.

### LINKED ISSUE #12: Author validation from DOI should be more stringent than last name
The author validation from the DOI is done by checking last names only. It is possible that there are multiple people with the same last name on a paper, so we should have a more stringent check.

kyleniemeyer:
[x] Fixes #11 
[x] Fixes #12 
[x] Tests added: `test_missing_author`, `test_extra_authors`, `test_two_authors_same_surname`

 - Validation checks that all authors given by DOI info are present
 - Validation checks that no extra authors are specified
 - Validation uses the `compare_name` function to more rigorously check author names against DOI info

@Niemeyer-Research-Group/ChemKED

### DIFF
--- a/pyked/tests/test_validation.py
+++ b/pyked/tests/test_validation.py
@@ -232,8 +232,8 @@ class TestValidator(object):
         """Ensure author validation can distinguish authors with same surname.
         """
         # missing Liuyan Lu from author list
-        authors = [{'name': 'Zhuyin Ren'}, {'name': 'Yufeng Liu'}},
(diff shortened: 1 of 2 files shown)


bryanwweber:
OK this seems fine to me and the tests pass, so I'm going to merge it so I can work on Cerberus 1.0

PR #28, opened 2016-11-21
Add mass fraction

### LINKED ISSUE #2: Allow specifying mass fraction for composition (needs Cerberus 1.0 release)
Currently only `mole-fraction` can be specified for the mixture `composition`. Using the new [`excludes` rule](http://docs.python-cerberus.org/en/latest/validation-rules.html#excludes) in [Cerberus 1.0](http://docs.python-cerberus.org/en/latest/changelog.html), we can allow the ChemKED file to specify one or the other:

```
composition: &composition
  type: list
  schema:
    type: dict
    schema:
      species:
        required: true
        type: string
      InChI:
        required: true
        type: string
      # Specify either mole fraction or mass fraction
      mole-fraction:
        required: true
        type: float
        min: 0.0
        max: 1.0
        excludes: mass-fraction
      mass-fraction:
        required: true
        type: float
        min: 0.0
        max: 1.0
        excludes: mole-fraction
```

### LINKED ISSUE #8: Decide how to handle species with fractional elemental representations and no InChI etc.
Some fuels don't have an InChI or similar identifier, such as global representations of practical fuels (e.g., C<sub>10.9</sub>H<sub>22.0</sub>). What should we do for the InChI for these fuels? One option is to put in an arbitrary string like `"This species doesn't have an InChI"` but that obviously won't validate if we get around to validating InChI values.

bryanwweber:
- [x] Fixes #2.
- [x] Closes #8.
- [x] Tests added

 - Add `mass-fraction` to `composition` schema
 - Add `mole-percent` to `composition` schema
 - Change `species` to `species-name` in `composition` schema
 - Enforce the minimum length of the composition list should be 1
 - Recommend the use of emoji in commit messages :tada:

This is 🚧  - work in progress - it needs tests still!

@Niemeyer-Research-Group/ChemKED

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,7 @@ You can find a [list of current issues](https://github.com/Niemeyer-Research-Gro
  * If you're unfamiliar with Pull Requests, please take a look at the [GitHub documentation for them](https://help.github.com/articles/proposing-changes-to-a-project-with-pull-requests/).
  * **Make sure the test suite passes** on your computer, and that test coverage doesn't go down. To do this, run `pytest -vv --cov=./` from the top-level directory.
  * *Always* add tests and docs for your code.
+ * The use of emoji in Pull Requests is encouraged with the format ":emoji: Commit summary". See [this list of suggested emoji.](https://github.com/slashsBin/styleguide-git-commit-message#suggested-emojis)
  * Please reference relevant GitHub issues in your commit messages using `GH123` or `#123`.
  * Changes should be [PEP8](https://www.python.org/dev/peps/pep-0008/) and [PEP257](https://www.python.org/dev/peps/pep-0257/) compatible.
  * Keep style fixes to a separate commit to make your pull request more readable.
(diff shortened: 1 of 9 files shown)


bryanwweber:
This is still 🚧 - I need to change the `cantera_composition` function to handle mass fraction and mole percent.

bryanwweber:
Apparently :sparkle: is different from :sparkles: 😦

bryanwweber:
OK I think this is all set now ✔️

kyleniemeyer:
Looks good; I think we need to add some tests related to specifying species with `atomic-structure`, but that can go with the changes needed to check InChI, SMILES, etc.

kyleniemeyer:
Before merging, can I ask why the decision to go with `atomic-structure` rather than `molecular-structure`?

bryanwweber:
`molecular-structure` implies some sort of species graph to me, whereas `atomic-structure` implies only the atoms that make up the molecule. I imagine in the future, we'll want to allow something like an RMG species graph, which makes more sense to me as a `molecular-structure`.

kyleniemeyer:
Ah, I see. So the difference is that right now we only give the makeup, and not the structure. Perhaps `atomic-composition` or `elemental-composition` would be better, then?

bryanwweber:
Why not both? :joy:

bryanwweber:
To document here, the `elemental-composition` and `atomic-composition` should be treated as synonyms, whenever we get around to processing them.

PR #66, opened 2017-07-19
Adds XML converters 

kyleniemeyer:
- [x] Tests added

 - Adds a YAML printer for ChemKED objects
 - Adds a function to convert ReSpecTh files to ChemKED files

@pr-omethe-us/chemked

### DIFF
--- a/docs/ck-tutorial.rst
+++ b/docs/ck-tutorial.rst
@@ -502,7 +502,7 @@ either from the command line
 
 .. code-block:: bash
 
-    python converter -i file.xml -o file.yaml
+    respth2ck -i file.xml -o file.yaml
 
 or via Python:
 
@@ -514,7 +514,7 @@ or via Python:
 Information about the person creating this new ChemKED file (e.g., file author name
 and their ORCID) may also be added via the ``file_author`` and ``file_author_orcid``
 arguments in Python or the corresponding command-line options `-fa` and `-fo`. More
-details can be found via ``python converter --help`` or
+details can be found via ``respth2ck --help`` or
 ``help(pyked.converters.ReSpecTh_to_ChemKED)``, respectively.
 
 PyKED also provides a converter to generate ReSpecTh files based on ChemKED records.
@@ -523,7 +523,7 @@ from the command line via
 
 .. code-block:: bash
 
-    python converter -i file.yaml -o file.xml
+    ck2respth -i file.yaml -o file.xml
 
 or from inside Python with
 
--- a/docs/ck-tutorial.rst
+++ b/docs/ck-tutorial.rst
@@ -490,6 +490,52 @@ shock tube at Stanford.
     ...
 
 
+Converting from ReSpecTh
+------------------------
+
+PyKED provides converter functions from (and to) the ReSpecTh XML format
+[Varga2015a]_, [Varga2015b]_, though ChemKED files created in this manner may
+require manual edits to fully satisfy our schema.
+
+Given a ReSpecTh XML file ``file.xml``, a corresponding ChemKED file can be created
+either from the command line
+
+.. code-block:: bash
+
+    python converter -i file.xml -o file.yaml
+
+or via Python:
+
+.. code-block:: Python
+
+    from pyked.converters import ReSpecTh_to_ChemKED
+    ReSpecTh_to_ChemKED('file.xml', 'file.yaml')
+
+Information about the person creating this new ChemKED file (e.g., file author name
+and their ORCID) may also be added via the ``file_author`` and ``file_author_orcid``
+arguments in Python or the corresponding command-line options `-fa` and `-fo`. More
+details can be found via ``python converter --help`` or
+``help(pyked.converters.ReSpecTh_to_ChemKED)``, respectively.
+
+PyKED also provides a converter to generate ReSpecTh files based on ChemKED records.
+Given a ChemKED file ``file.yaml``, a corresponding ReSpecTh file can be created
+from the command line via
+
+.. code-block:: bash
+
+    python converter -i file.yaml -o file.xml
+
+or from inside Python with
+
+.. code-block:: Python
+
+    from pyked import ChemKED
+    c = ChemKED('file.yaml')
+    c.convert_to_ReSpecTh('file.xml')
+
+Note that some information, or granularity of details, may be lost in this conversion.
+
+
 Works Cited
 -----------
 
@@ -506,3 +552,11 @@ Works Cited
                  Ignition Delay Times for the Butanol Isomers." *Combustion and Flame*
                  159 (2): 516–27.
                  doi:\ `10.1016/j.combustflame.2011.08.014 <https://doi.org/10.1016/j.combustflame.2011.08.014>`__.
+
+.. [Varga2015a] Varga, Tamás, Tamás Turányi, Eszter Czinki, Tibor Furtenbacher, Attila G. Császár,
+                "ReSpecTh: a joint reaction kinetics, spectroscopy, and thermochemistry information
+                system." 2015. Proceedings of the 7th European Combustion Meeting,
+                Budapest, Hungary.
+
+.. [Varga2015b] Varga, Tamás. "ReSpecTh Kinetics Data Format Specification v1.0." 2015.
+                http://respecth.hu/
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -337,7 +337,7 @@ class ChemKED(object):
                 component = etree.SubElement(prop, 'component')
                 species_link = etree.SubElement(component, 'speciesLink')
                 species_link.set('preferredKey', species['species-name'])
-                if species.get('InChI'):
+                if species.get('InChI') is not None
                     species_link.set('InChI', species['InChI'])
 
                 amount = etree.SubElement(component, 'amount')
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -355,18 +355,19 @@ class ChemKED(object):
         if len(self.datapoints) > 1:
             for prop_name in datagroup_properties:
                 attribute = prop_name.replace(' ', '_')
-                quantity = getattr(self.datapoints[0], attribute)
-                if (quantity != None and
-                    all([quantity == getattr(dp, attribute) for dp in self.datapoints])
-                    ):
+                quantities = [getattr(dp, attribute, False) for dp in self.datapoints]
+
+                # All quantities must have the property in question and all the
+                # values must be equal
+                if all(quantities) and quantities.count(quantities[0]) == len(quantities):
                     common.append(prop_name)
                     prop = etree.SubElement(common_properties, 'property')
                     prop.set('description', '')
                     prop.set('name', prop_name)
-                    prop.set('units', str(quantity.units))
+                    prop.set('units', str(quantity[0].units))
 
                     value = etree.SubElement(prop, 'value')
-                    value.text = str(quantity.magnitude)
+                    value.text = str(quantity[0].magnitude)
 
         # Ignition delay can't be common, unless only a single datapoint.
 
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -424,13 +424,12 @@ class ChemKED(object):
                             value.text = str(item['amount'].magnitude)
 
         # if RCM and has volume history, need a second dataGroup
-        if (len(self.datapoints) > 1 and
-            any([getattr(dp, 'volume_history', None) for dp in self.datapoints])
-            ):
+        has_volume_history = any([hasattr(dp, 'volume_history') for dp in self.datapoints])
+        if len(self.datapoints) > 1 and has_volume_history:
             raise NotImplementedError('Error: ReSpecTh files do not support multiple datapoints '
                                       'with a volume history.'
                                       )
-            # TODO: what if they share the same history? Does this happen?
+                                      
         elif getattr(self.datapoints[0], 'volume_history', None):
             datagroup = etree.SubElement(root, 'dataGroup')
             datagroup.set('id', 'dg1')
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -409,7 +409,14 @@ class ChemKED(object):
             datapoint = etree.SubElement(datagroup, 'dataPoint')
             for idx in property_idx:
                 value = etree.SubElement(datapoint, idx)
-                value.text = str(getattr(dp, property_idx[idx].replace(' ', '_')).magnitude)
+                # handle regular properties a bit differently than composition
+                if property_idx[idx] in datagroup_properties:
+                    value.text = str(getattr(dp, property_idx[idx].replace(' ', '_')).magnitude)
+                else:
+                    # composition
+                    for item in dp.composition:
+                        if item['species-name'] == property_idx[idx]:
+                            value.text = str(item['amount'].magnitude)
 
         # if RCM and has volume history, need a second dataGroup
         if (len(self.datapoints) > 1 and
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -325,8 +325,14 @@ class ChemKED(object):
         # among datapoints that tend to be common
         common = []
         composition = self.datapoints[0].composition
+
         # Composition type *has* to be the same
         composition_type = self.datapoints[0].composition_type
+        if not all(dp.composition_type == composition_type for dp in self.datapoints):
+            raise NotImplementedError('Error: ReSpecTh does not support varying composition '
+                                      'type among datapoints.'
+                                      )
+
         if all([composition == dp.composition for dp in self.datapoints]):
             # initial composition is common
             common.append('composition')
@@ -337,7 +343,7 @@ class ChemKED(object):
                 component = etree.SubElement(prop, 'component')
                 species_link = etree.SubElement(component, 'speciesLink')
                 species_link.set('preferredKey', species['species-name'])
-                if species.get('InChI') is not None
+                if species.get('InChI') is not None:
                     species_link.set('InChI', species['InChI'])
 
                 amount = etree.SubElement(component, 'amount')
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -464,7 +464,7 @@ class ChemKED(object):
 
         et = etree.ElementTree(root)
         et.write(filename, encoding='utf-8', xml_declaration=True)
-        print('Converted to ' + filename_xml)
+        print('Converted to ' + filename)
 
 
 class DataPoint(object):
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -364,10 +364,10 @@ class ChemKED(object):
                     prop = etree.SubElement(common_properties, 'property')
                     prop.set('description', '')
                     prop.set('name', prop_name)
-                    prop.set('units', str(quantity[0].units))
+                    prop.set('units', str(quantities[0].units))
 
                     value = etree.SubElement(prop, 'value')
-                    value.text = str(quantity[0].magnitude)
+                    value.text = str(quantities[0].magnitude)
 
         # Ignition delay can't be common, unless only a single datapoint.
 
@@ -429,25 +429,26 @@ class ChemKED(object):
         for dp in self.datapoints:
             datapoint = etree.SubElement(datagroup, 'dataPoint')
             for idx, val in property_idx.items():
-                value = etree.SubElement(datapoint, idx)
                 # handle regular properties a bit differently than composition
                 if val['name'] in datagroup_properties:
+                    value = etree.SubElement(datapoint, idx)
                     quantity = getattr(dp, val['name'].replace(' ', '_')).to(val['units'])
                     value.text = str(quantity.magnitude)
                 else:
                     # composition
                     for item in dp.composition:
                         if item['species-name'] == val['name']:
+                            value = etree.SubElement(datapoint, idx)
                             value.text = str(item['amount'].magnitude)
 
         # if RCM and has volume history, need a second dataGroup
-        has_volume_history = any([hasattr(dp, 'volume_history') for dp in self.datapoints])
+        has_volume_history = any([dp.volume_history is not None for dp in self.datapoints])
         if len(self.datapoints) > 1 and has_volume_history:
             raise NotImplementedError('Error: ReSpecTh files do not support multiple datapoints '
                                       'with a volume history.'
                                       )
 
-        elif hasattr(self.datapoints[0], 'volume_history'):
+        elif self.datapoints[0].volume_history is not None:
             datagroup = etree.SubElement(root, 'dataGroup')
             datagroup.set('id', 'dg2')
             datagroup_link = etree.SubElement(datagroup, 'dataGroupLink')
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -432,7 +432,7 @@ class ChemKED(object):
 
         elif hasattr(self.datapoints[0], 'volume_history'):
             datagroup = etree.SubElement(root, 'dataGroup')
-            datagroup.set('id', 'dg1')
+            datagroup.set('id', 'dg2')
             datagroup_link = etree.SubElement(datagroup, 'dataGroupLink')
             datagroup_link.set('dataGroupID', '')
             datagroup_link.set('dataPointID', '')
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -6,12 +6,14 @@ from os.path import exists
 from collections import namedtuple
 from warnings import warn
 from copy import deepcopy
+import xml.etree.ElementTree as etree
 
 import numpy as np
 
 # Local imports
 from .validation import schema, OurValidator, yaml
 from .utils import Q_
+from .converters import datagroup_properties
 
 VolumeHistory = namedtuple('VolumeHistory', ['time', 'volume'])
 VolumeHistory.__doc__ = 'Time history of the volume in an RCM experiment'
@@ -45,7 +47,7 @@ class ChemKED(object):
 
     Args:
         yaml_file (`str`, optional): The filename of the YAML database in ChemKED format.
-        dict_input (`str`, optional): A dictionary with the parsed ouput of YAML file in ChemKED
+        dict_input (`dict`, optional): A dictionary with the parsed ouput of YAML file in ChemKED
             format.
         skip_validation (`bool`, optional): Whether validation of the ChemKED should be done. Must
             be supplied as a keyword-argument.
@@ -262,6 +264,207 @@ class ChemKED(object):
 
         with open(filename, 'w') as yaml_file:
             yaml.dump(self._properties, yaml_file)
+
+    def convert_to_ReSpecTh(self, filename):
+        """Convert ChemKED record to ReSpecTh XML file.
+
+        This converter uses common information in a ChemKED file to generate a
+        ReSpecTh XML file. Note that some information may be lost, as ChemKED stores
+        some additional attributes.
+
+        Arguments:
+            filename (`str`): Filename for output ReSpecTh XML file.
+
+        Example:
+            >>> dataset = ChemKED(yaml_file)
+            >>> dataset.convert_to_ReSpecTh(xml_file)
+        """
+        root = etree.Element('experiment')
+
+        file_author = etree.SubElement(root, 'fileAuthor')
+        file_author.text = self.file_author['name']
+
+        # right now ChemKED just uses an integer file version
+        file_version = etree.SubElement(root, 'fileVersion')
+        major_version = etree.SubElement(file_version, 'major')
+        major_version.text = str(self.file_version)
+        minor_version = etree.SubElement(file_version, 'minor')
+        minor_version.text = '0'
+
+        respecth_version = etree.SubElement(root, 'ReSpecThVersion')
+        major_version = etree.SubElement(respecth_version, 'major')
+        major_version.text = '1'
+        minor_version = etree.SubElement(respecth_version, 'minor')
+        minor_version.text = '0'
+
+        # Only ignition delay currently supported
+        exp = etree.SubElement(root, 'experimentType')
+        if self.experiment_type == 'ignition delay':
+            exp.text = 'Ignition delay measurement'
+        else:
+            raise NotImplementedError('Only ignition delay type supported for conversion.')
+
+        reference = etree.SubElement(root, 'bibliographyLink')
+        citation = ''
+        for author in self.reference.authors:
+            citation += author['name'] + ', '
+        citation += (self.reference.journal + ' (' + str(self.reference.year) + ') ' +
+                     str(self.reference.volume) + ':' + self.reference.pages + '. ' +
+                     self.reference.detail
+                     )
+        reference.set('preferredKey', citation)
+        reference.set('doi', self.reference.doi)
+
+        apparatus = etree.SubElement(root, 'apparatus')
+        kind = etree.SubElement(apparatus, 'kind')
+        kind.text = self.apparatus.kind
+
+        common_properties = etree.SubElement(root, 'commonProperties')
+        # ChemKED objects have no common properties once loaded. Check for properties
+        # among datapoints that tend to be common
+        common = []
+        composition = self.datapoints[0].composition
+        # Composition type *has* to be the same
+        composition_type = self.datapoints[0].composition_type
+        if all([composition == dp.composition for dp in self.datapoints]):
+            # initial composition is common
+            common.append('composition')
+            prop = etree.SubElement(common_properties, 'property')
+            prop.set('name', 'initial composition')
+
+            for species in composition:
+                component = etree.SubElement(prop, 'component')
+                species_link = etree.SubElement(component, 'speciesLink')
+                species_link.set('preferredKey', species['species-name'])
+                if species.get('InChI'):
+                    species_link.set('InChI', species['InChI'])
+
+                amount = etree.SubElement(component, 'amount')
+                amount.set('units', composition_type)
+                amount.text = str(species['amount'].magnitude)
+
+        # If multiple datapoints present, then find any common properties. If only
+        # one datapoint, then composition should be the only "common" property.
+        if len(self.datapoints) > 1:
+            for prop_name in datagroup_properties:
+                attribute = prop_name.replace(' ', '_')
+                quantity = getattr(self.datapoints[0], attribute)
+                if (quantity != None and
+                    all([quantity == getattr(dp, attribute) for dp in self.datapoints])
+                    ):
+                    common.append(prop_name)
+                    prop = etree.SubElement(common_properties, 'property')
+                    prop.set('description', '')
+                    prop.set('name', prop_name)
+                    prop.set('units', str(quantity.units))
+
+                    value = etree.SubElement(prop, 'value')
+                    value.text = str(quantity.magnitude)
+
+        # Ignition delay can't be common, unless only a single datapoint.
+
+        datagroup = etree.SubElement(root, 'dataGroup')
+        datagroup.set('id', 'dg1')
+        datagroup_link = etree.SubElement(datagroup, 'dataGroupLink')
+        datagroup_link.set('dataGroupID', '')
+        datagroup_link.set('dataPointID', '')
+
+        property_idx = {}
+        labels = {'temperature': 'T', 'pressure': 'P',
+                  'ignition delay': 'tau', 'pressure rise': 'dP/dt',
+                  }
+
+        for prop_name in datagroup_properties:
+            attribute = prop_name.replace(' ', '_')
+            if (prop_name not in common and
+                any([getattr(dp, attribute, None) for dp in self.datapoints])
+                ):
+                prop = etree.SubElement(datagroup, 'property')
+                prop.set('description', '')
+                prop.set('name', prop_name)
+                prop.set('units', str(getattr(self.datapoints[0], attribute).units))
+                idx = 'x{}'.format(len(property_idx) + 1)
+                property_idx[idx] = prop_name
+                prop.set('id', idx)
+                prop.set('label', labels[prop_name])
+
+        if 'composition' not in common:
+            for species in self.datapoints[0].composition:
+                prop = etree.SubElement(datagroup, 'property')
+                prop.set('description', '')
+
+                idx = 'x{}'.format(len(property_idx) + 1)
+                property_idx[idx] = species['species-name']
+                prop.set('id', idx)
+                prop.set('label', '[' + species['species-name'] + ']')
+                prop.set('name', 'composition')
+                prop.set('units', self.datapoints[0].composition_type)
+
+                species_link = etree.SubElement(prop, 'speciesLink')
+                species_link.set('preferredKey', species['species-name'])
+                if species.get('InChI'):
+                    species_link.set('InChI', species['InChI'])
+
+        for dp in self.datapoints:
+            datapoint = etree.SubElement(datagroup, 'dataPoint')
+            for idx in property_idx:
+                value = etree.SubElement(datapoint, idx)
+                value.text = str(getattr(dp, property_idx[idx].replace(' ', '_')).magnitude)
+
+        # if RCM and has volume history, need a second dataGroup
+        if (len(self.datapoints) > 1 and
+            any([getattr(dp, 'volume_history', None) for dp in self.datapoints])
+            ):
+            raise NotImplementedError('Error: ReSpecTh files do not support multiple datapoints '
+                                      'with a volume history.'
+                                      )
+            # TODO: what if they share the same history? Does this happen?
+        elif getattr(self.datapoints[0], 'volume_history', None):
+            datagroup = etree.SubElement(root, 'dataGroup')
+            datagroup.set('id', 'dg1')
+            datagroup_link = etree.SubElement(datagroup, 'dataGroupLink')
+            datagroup_link.set('dataGroupID', '')
+            datagroup_link.set('dataPointID', '')
+
+            # Volume history has two properties: time and volume.
+            volume_history = self.datapoints[0].volume_history
+            prop = etree.SubElement(datagroup, 'property')
+            prop.set('description', '')
+            prop.set('name', 'time')
+            prop.set('units', str(volume_history.time.units))
+            time_idx = 'x{}'.format(len(property_idx) + 1)
+            prop.set('id', time_idx)
+            prop.set('label', 't')
+
+            prop = etree.SubElement(datagroup, 'property')
+            prop.set('description', '')
+            prop.set('name', 'volume')
+            prop.set('units', str(volume_history.volume.units))
+            volume_idx = 'x{}'.format(len(property_idx) + 2)
+            prop.set('id', volume_idx)
+            prop.set('label', 'V')
+
+            for time, volume in zip(volume_history.time, volume_history.volume):
+                datapoint = etree.SubElement(datagroup, 'dataPoint')
+                value = etree.SubElement(datapoint, time_idx)
+                value.text = str(time.magnitude)
+                value = etree.SubElement(datapoint, volume_idx)
+                value.text = str(volume.magnitude)
+
+        # In ReSpecTh files all datapoints share ignition type
+        ignition = etree.SubElement(root, 'ignitionType')
+        if self.datapoints[0].ignition_type['target'] == 'pressure':
+            ignition.set('target', 'P')
+        elif self.datapoints[0].ignition_type['target'] == 'temperature':
+            ignition.set('target', 'T')
+        else:
+            # options left are species
+            ignition.set('target', self.datapoints[0].ignition_type['target'])
+        ignition.set('type', self.datapoints[0].ignition_type['type'])
+
+        et = etree.ElementTree(root)
+        et.write(filename, encoding='utf-8', xml_declaration=True)
+        print('Converted to ' + filename_xml)
 
 
 class DataPoint(object):
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -384,33 +384,42 @@ class ChemKED(object):
 
         for prop_name in datagroup_properties:
             attribute = prop_name.replace(' ', '_')
-            if (prop_name not in common and
-                any([getattr(dp, attribute, None) for dp in self.datapoints])
-                ):
-                prop = etree.SubElement(datagroup, 'property')
-                prop.set('description', '')
-                prop.set('name', prop_name)
-                prop.set('units', str(getattr(self.datapoints[0], attribute).units))
-                idx = 'x{}'.format(len(property_idx) + 1)
-                property_idx[idx] = prop_name
-                prop.set('id', idx)
-                prop.set('label', labels[prop_name])
+            # This can't be hasattr because properties are set to the value None
+            # if no value is specified in the file, so the attribute always exists
+            prop_indices = [i for i, dp in enumerate(self.datapoints)
+                            if getattr(dp, attribute) is not None
+                            ]
+            if prop_name in common or not prop_indices:
+                continue
+
+            prop = etree.SubElement(datagroup, 'property')
+            prop.set('description', '')
+            prop.set('name', prop_name)
+            units = str(getattr(self.datapoints[prop_indices[0]], attribute).units)
+            prop.set('units', units)
+            idx = 'x{}'.format(len(property_idx) + 1)
+            property_idx[idx] = {'name': prop_name, 'units': units}
+            prop.set('id', idx)
+            prop.set('label', labels[prop_name])
 
         # Need to handle datapoints with possibly different species in the initial composition
         if 'composition' not in common:
             for dp in self.datapoints:
                 for species in dp.composition:
                     # Only add new property for species not already considered
-                    if species['species-name'] not in property_idx.values():
+                    has_spec = any([species['species-name'] in d.values()
+                                    for d in property_idx.values()
+                                    ])
+                    if not has_spec:
                         prop = etree.SubElement(datagroup, 'property')
                         prop.set('description', '')
 
                         idx = 'x{}'.format(len(property_idx) + 1)
-                        property_idx[idx] = species['species-name']
+                        property_idx[idx] = {'name': species['species-name']}
                         prop.set('id', idx)
                         prop.set('label', '[' + species['species-name'] + ']')
                         prop.set('name', 'composition')
-                        prop.set('units', dp.composition_type)
+                        prop.set('units', self.datapoints[0].composition_type)
 
                         species_link = etree.SubElement(prop, 'speciesLink')
                         species_link.set('preferredKey', species['species-name'])
@@ -419,15 +428,16 @@ class ChemKED(object):
 
         for dp in self.datapoints:
             datapoint = etree.SubElement(datagroup, 'dataPoint')
-            for idx in property_idx:
+            for idx, val in property_idx.items():
                 value = etree.SubElement(datapoint, idx)
                 # handle regular properties a bit differently than composition
-                if property_idx[idx] in datagroup_properties:
-                    value.text = str(getattr(dp, property_idx[idx].replace(' ', '_')).magnitude)
+                if val['name'] in datagroup_properties:
+                    quantity = getattr(dp, val['name'].replace(' ', '_')).to(val['units'])
+                    value.text = str(quantity.magnitude)
                 else:
                     # composition
                     for item in dp.composition:
-                        if item['species-name'] == property_idx[idx]:
+                        if item['species-name'] == val['name']:
                             value.text = str(item['amount'].magnitude)
 
         # if RCM and has volume history, need a second dataGroup
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -429,8 +429,8 @@ class ChemKED(object):
             raise NotImplementedError('Error: ReSpecTh files do not support multiple datapoints '
                                       'with a volume history.'
                                       )
-                                      
-        elif getattr(self.datapoints[0], 'volume_history', None):
+
+        elif hasattr(self.datapoints[0], 'volume_history'):
             datagroup = etree.SubElement(root, 'dataGroup')
             datagroup.set('id', 'dg1')
             datagroup_link = etree.SubElement(datagroup, 'dataGroupLink')
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -258,9 +258,9 @@ class ChemKED(object):
         """
         # Ensure file isn't already present
         if exists(filename) and not overwrite:
-            raise NameError(filename + ' already present. Specify "overwrite=True" '
-                            'to overwrite, or rename.'
-                            )
+            raise OSError(filename + ' already present. Specify "overwrite=True" '
+                          'to overwrite, or rename.'
+                          )
 
         with open(filename, 'w') as yaml_file:
             yaml.dump(self._properties, yaml_file)
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -7,6 +7,7 @@ from collections import namedtuple
 from warnings import warn
 from copy import deepcopy
 import xml.etree.ElementTree as etree
+import xml.dom.minidom as minidom
 
 import numpy as np
 
@@ -388,22 +389,26 @@ class ChemKED(object):
                 prop.set('id', idx)
                 prop.set('label', labels[prop_name])
 
+        # Need to handle datapoints with possibly different species in the initial composition
         if 'composition' not in common:
-            for species in self.datapoints[0].composition:
-                prop = etree.SubElement(datagroup, 'property')
-                prop.set('description', '')
+            for dp in self.datapoints:
+                for species in dp.composition:
+                    # Only add new property for species not already considered
+                    if species['species-name'] not in property_idx.values():
+                        prop = etree.SubElement(datagroup, 'property')
+                        prop.set('description', '')
 
-                idx = 'x{}'.format(len(property_idx) + 1)
-                property_idx[idx] = species['species-name']
-                prop.set('id', idx)
-                prop.set('label', '[' + species['species-name'] + ']')
-                prop.set('name', 'composition')
-                prop.set('units', self.datapoints[0].composition_type)
+                        idx = 'x{}'.format(len(property_idx) + 1)
+                        property_idx[idx] = species['species-name']
+                        prop.set('id', idx)
+                        prop.set('label', '[' + species['species-name'] + ']')
+                        prop.set('name', 'composition')
+                        prop.set('units', dp.composition_type)
 
-                species_link = etree.SubElement(prop, 'speciesLink')
-                species_link.set('preferredKey', species['species-name'])
-                if species.get('InChI'):
-                    species_link.set('InChI', species['InChI'])
+                        species_link = etree.SubElement(prop, 'speciesLink')
+                        species_link.set('preferredKey', species['species-name'])
+                        if species.get('InChI'):
+                            species_link.set('InChI', species['InChI'])
 
         for dp in self.datapoints:
             datapoint = etree.SubElement(datagroup, 'dataPoint')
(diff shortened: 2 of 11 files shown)


kyleniemeyer:
There are currently two XML testfiles, but no actual tests to check the converter.

pyked/chemked.py
```diff
@@ -175,6 +175,7 @@ def get_dataframe(self, output_columns=None):
 
         valid_labels = [a.replace('_', ' ') for a in self.__dict__ if not a.startswith('__')]
         valid_labels.remove('datapoints')
+        valid_labels.remove(' properties')
```
bryanwweber: I would suggest moving this a few lines up and adding it to the list generator,

    valid_labels = [a.replace('_', ' ') for a in self.__dict__ if not (a.startswith('__') or a.startswith('_'))]

That way we can add other "private" properties

kyleniemeyer: Sounds good—this is addressed in 0628977

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        reference['doi'] = elem.attrib['doi']
+        ref = None
+        try:
+            ref = habanero.Crossref().works(ids=reference['doi'])['message']
+        except (HTTPError, habanero.RequestError):
+            print('DOI not found')
+            raise KeyError
+        # TODO: remove UnboundLocalError after habanero fixed
+        except (ConnectionError, UnboundLocalError):
+            warn('network not available, DOI not validated.')
```
bryanwweber: Is this the appropriate warning message for this spot? Same comment before about (ab)using `KeyError` here.

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    except KeyError:
+        print('Warning: missing doi attribute in bibliographyLink')
+        print('Setting "detail" key as a fallback; please update.')
+        try:
+            reference['detail'] = elem.attrib['preferredKey']
+            if reference['detail'][-1] != '.':
+                reference['detail'] += '.'
+        except KeyError:
+            # Need one of DOI or preferredKey
+            raise KeyError('Error: missing preferredKey attribute in bibliographyLink')
```
bryanwweber: Should this be a `MissingElementError`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+    Returns:
+        properties (`dict`): Dictionary with experiment type and apparatus information.
+    """
+    properties = {}
+    if root.find('experimentType').text == 'Ignition delay measurement':
+        properties['experiment-type'] = 'ignition delay'
+    else:
+        #TODO: support additional experimentTypes
+        raise KeyError('experimentType not ignition delay measurement')
```
bryanwweber: `NotImplementedError`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+    properties['apparatus'] = {'kind': '', 'institution': '', 'facility': ''}
```
bryanwweber: Can you fill in the `institution` and `facility` from the ReSpecTh file?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+            for child in elem.iter('component'):
+                spec = {}
+                spec['species-name'] = child.find('speciesLink').attrib['preferredKey']
+
+                # use InChI for unique species identifier (if present)
+                try:
+                    spec['InChI'] = child.find('speciesLink').attrib['InChI']
+                except KeyError:
+                    # TODO: add InChI validator/search
+                    print('Warning: missing InChI for species ' + spec['species-name'])
```
bryanwweber: Logging?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+            quantity = 1.0 * unit_registry(units)
+            try:
+                quantity.to(property_units[field])
+            except pint.DimensionalityError:
+                raise KeyError('Error: units incompatible for property ' + name)
+
+            properties[field] = [' '.join([elem.find('value').text, units])]
+
+        else:
+            raise KeyError('Property ' + name + ' not supported as common property.')
```
bryanwweber: Probably we should just `warn` here, rather than bombing out, since its likely that people will run into this, but its not a critical problem, is it?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        elif name in ['temperature', 'pressure', 'pressure rise', 'compression time']:
+            field = name.replace(' ', '-')
+            units = elem.attrib['units']
+            if units == 'Torr':
+                units = 'torr'
+            quantity = 1.0 * unit_registry(units)
+            try:
+                quantity.to(property_units[field])
+            except pint.DimensionalityError:
+                raise KeyError('Error: units incompatible for property ' + name)
```
bryanwweber: I don't think `KeyError` is the right one to use here. Maybe `KeywordError`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        root (`etree.Element`): Root of ReSpecTh XML file
+
+    Returns:
+        properties (`dict`): Dictionary with ignition type/target information
+    """
+    ignition = {}
+    elem = root.find('ignitionType')
+
+    if elem is None:
+        raise AttributeError('missing ignitionType')
```
bryanwweber: `MissingElementError`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+    try:
```
bryanwweber: Consider using the `get` method of Python dictionaries to avoid having to catch `KeyErrors`? Also, these should probably be `MissingElementErrors`

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    elif ign_target == 'CHEX':
+        ign_target = 'CH*'
+
+    if ign_target not in ['P', 'T', 'OH', 'OH*', 'CH*', 'CH']:
+        raise KeyError(ign_target + ' not valid ignition target')
+
+    if ign_type not in ['max', 'd/dt max', '1/2 max', 'min']:
+        raise KeyError(ign_type + ' not valid ignition type')
+
+    if ign_target == 'P':
```
bryanwweber: Why not put all the `ign_target` conversions into one `if-elif` block?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        root (`etree.Element`): Root of ReSpecTh XML file
+
+    Returns:
+        properties (`dict`): Dictionary with ignition delay data
+    """
+    # Shock tube experiment will have one data group, while RCM may have one
+    # or two (one for ignition delay, one for volume-history)
+    dataGroups = root.findall('dataGroup')
+    if not dataGroups:
+        raise AttributeError('Error: at least one dataGroup needed.')
```
bryanwweber: `MissingElementError`? Also, will it be `False`,  `None`, or an empty list? If `None`, should be `if dataGroups is not None:`

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    unit_id = {}
+    # get properties of dataGroup
+    for prop in dataGroup.findall('property'):
+        unit_id[prop.attrib['id']] = prop.attrib['units']
+        property_id[prop.attrib['id']] = prop.attrib['name'].replace(' ', '-')
+        if property_id[prop.attrib['id']] not in [
+            'temperature', 'pressure', 'ignition-delay', 'pressure-rise'
+            ]:
+            raise KeyError(property_id[prop.attrib['id']] + ' not valid dataPoint property')
+    if property_id == {}:
```
bryanwweber: Can do `if not property_id:`, https://docs.python.org/3/library/stdtypes.html#truth-value-testing

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    property_id = {}
+    unit_id = {}
+    # get properties of dataGroup
+    for prop in dataGroup.findall('property'):
+        unit_id[prop.attrib['id']] = prop.attrib['units']
+        property_id[prop.attrib['id']] = prop.attrib['name'].replace(' ', '-')
+        if property_id[prop.attrib['id']] not in [
+            'temperature', 'pressure', 'ignition-delay', 'pressure-rise'
+            ]:
+            raise KeyError(property_id[prop.attrib['id']] + ' not valid dataPoint property')
```
bryanwweber: `KeywordError`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    if property_id == {}:
+        raise AttributeError('Error: at least one property needed in dataGroup.')
```
bryanwweber: `MissingElementError`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    # all situations will have main experimental data in first dataGroup
+    dataGroup = dataGroups[0]
+    property_id = {}
+    unit_id = {}
+    # get properties of dataGroup
+    for prop in dataGroup.findall('property'):
+        unit_id[prop.attrib['id']] = prop.attrib['units']
+        property_id[prop.attrib['id']] = prop.attrib['name'].replace(' ', '-')
+        if property_id[prop.attrib['id']] not in [
+            'temperature', 'pressure', 'ignition-delay', 'pressure-rise'
```
bryanwweber: It might be more convenient for later extension if you define this as a separate list outside the `for` loop.

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        raise AttributeError('Error: at least one dataGroup needed.')
+
+    # all situations will have main experimental data in first dataGroup
+    dataGroup = dataGroups[0]
+    property_id = {}
+    unit_id = {}
+    # get properties of dataGroup
+    for prop in dataGroup.findall('property'):
+        unit_id[prop.attrib['id']] = prop.attrib['units']
+        property_id[prop.attrib['id']] = prop.attrib['name'].replace(' ', '-')
```
bryanwweber: I'm not totally sure, but I think it would be better to assign to a temporary variable here, so that the dictionary is not filled with invalid values (if, for instance, someone catches `KeyError`s and passes on them in a code that calls this one). Not totally sure if that's a valid concern, but figured I'd raise it.

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        datapoint = {}
+        for val in dp:
+            units = unit_id[val.tag]
+            if units == 'Torr':
+                units = 'torr'
+            datapoint[property_id[val.tag]] = [val.text + ' ' + units]
+        datapoints.append(datapoint)
+
+    if len(datapoints) == 0:
+        raise AttributeError('Error: need at least one dataPoint.')
```
bryanwweber: `MissingElementError`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+def convert_ReSpecTh(filename_xml, output='', file_author='', file_author_orcid=''):
+    """Convert ReSpecTh XML file to ChemKED YAML file.
+
+    Args:
+        filename_xml (`str`): Name of ReSpecTh XML file to be converted.
+        output (`str`, optional): Output path for converted file.
+        file_author (`str`, optional): Name to override original file author
+        file_author_orcid (`str`, optional): ORCID of file author
+
+    Returns:
```
bryanwweber: I don't think its typical for functions like this to return the filename, unless its intended to be used for further processing. Since the purpose of this function is to write the output file, I'd say don't return anything.

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        filename_yaml (`str`): Name of newly created ChemKED YAML file.
+    """
+    assert os.path.isfile(filename_xml), 'Error: ' + filename_xml + ' file missing'
+
+    # get all information from XML file
+    try:
+        tree = etree.parse(filename_xml)
+    except OSError:
```
bryanwweber: Why is this `try` necessary? Is the `OSError` default message not clear? (I'm really not sure what the default message looks like :smile:)

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+    # get file metadata
+    properties = get_file_metadata(root)
+
+    # get reference info
+    properties['reference'] = get_reference(root)
+    # Save name of original data filename
+    if properties['reference'].get('detail') is None:
+        properties['reference']['detail'] = ''
+    properties['reference']['detail'] += ('Converted from XML file ' +
```
bryanwweber: `Converted from ReSpecTh XML file...`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    # Get properties shared across the file
+    properties['common-properties'] = get_common_properties(root)
+
+    # Determine definition of ignition delay
+    properties['common-properties']['ignition-type'] = get_ignition_type(root)
+
+    # Now parse ignition delay datapoints
+    properties['datapoints'] = get_datapoints(root)
+
+    # Get compression time for RCM, if volume history given
```
bryanwweber: I would suggest not to include commented out code

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        raise KeywordError('Time values needed for volume history')
+    elif (any(['volume' in dp for dp in properties['datapoints']]) and
+          not any(['time' in dp for dp in properties['datapoints']])
+          ):
+        raise KeywordError('Time values needed for volume history')
+
+    if ('compression-time' in properties['common-properties'] or
+        any([dp for dp in properties['datapoints'] if dp.get('compression-time')])
+        ) and properties['apparatus']['kind'] == 'shock tube':
+        raise KeyError('Compression time cannot be defined for shock tube.')
```
bryanwweber: `KeywordError`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+    if ('pressure-rise' in properties['common-properties'] or
+        any([dp for dp in properties['datapoints'] if dp.get('pressure-rise')])
+        ) and properties['apparatus']['kind'] == 'rapid compression machine':
+        raise KeyError('Pressure rise cannot be defined for RCM.')
```
bryanwweber: `KeywordError`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    if file_author_orcid:
+        properties['file-author']['ORCID'] = file_author_orcid
+
+    # Now go through datapoints and apply common properties
+    for idx in range(len(properties['datapoints'])):
+        for prop in properties['common-properties']:
+            properties['datapoints'][idx][prop] = properties['common-properties'][prop]
+
+    # compression time doesn't belong in common-properties
+    properties['common-properties'].pop('compression-time', None)
```
bryanwweber: How did `compression-time` get into the `common-properties`? Doesn't the code check for invalid common properties in that function?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+    filename_yaml = os.path.splitext(os.path.basename(filename_xml))[0] + '.yaml'
+
+    # add path
+    filename_yaml = os.path.join(output, filename_yaml)
+
+    with open(filename_yaml, 'w') as outfile:
+        outfile.write(yaml.dump(properties, default_flow_style=False))
+    print('Converted to ' + filename_yaml)
```
bryanwweber: Logging?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+        datapoints[0]['volume-history'] = volume_history
+
+    elif len(dataGroups) > 2:
+        raise NotImplementedError('More than two DataGroups not supported.')
+
+    return datapoints
+
+
+def convert_ReSpecTh(filename_xml, output='', file_author='', file_author_orcid=''):
```
bryanwweber: Should we offer a `filename_ck` argument if people want to change the filename of the output?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        for prop in properties['common-properties']:
+            properties['datapoints'][idx][prop] = properties['common-properties'][prop]
+
+    # compression time doesn't belong in common-properties
+    properties['common-properties'].pop('compression-time', None)
+
+    filename_yaml = os.path.splitext(os.path.basename(filename_xml))[0] + '.yaml'
+
+    # add path
+    filename_yaml = os.path.join(output, filename_yaml)
```
bryanwweber: Change to `filename_ck`?

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+                 )
+    reference.set('preferredKey', citation)
+    reference.set('doi', c.reference.doi)
+
+    apparatus = etree.SubElement(root, 'apparatus')
+    kind = etree.SubElement(apparatus, 'kind')
+    kind.text = c.apparatus.kind
+
+    common_properties = etree.SubElement(root, 'commonProperties')
+    # ChemKED objects have no common properties once loaded... can we check for properties
```
bryanwweber: I don't think we need to check... the point of common properties is to ensure that values are consistent, but they're guaranteed to be consistent by using the script here so it shouldn't be a problem.

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    datagroup_link = etree.SubElement(datagroup, 'dataGroupLink')
+    datagroup_link.set('dataGroupID', '')
+    datagroup_link.set('dataPointID', '')
+
+    property_idx = {}
+    labels = {'temperature': 'T', 'pressure': 'P',
+              'ignition delay': 'tau', 'pressure rise': 'dP/dt',
+              }
+
+    for prop_name in ['temperature', 'pressure', 'ignition delay', 'pressure rise']:
```
bryanwweber: Consider storing the `prop_name`s in a separate list, for easy future extensibility

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+            value.text = str(getattr(dp, property_idx[idx].replace(' ', '_')).magnitude)
+
+    # if RCM and has volume history, need a second dataGroup
+    if (len(c.datapoints) > 1 and
+        any([getattr(dp, 'volume_history', None) for dp in c.datapoints])
+        ):
+        raise NotImplementedError('Error: ReSpecTh files do not support multiple datapoints '
+                                  'with a volume history.'
+                                  )
+        # TODO: what if they share the same history? Does this happen?
```
bryanwweber: Not in my experience, unless someone is taking a shortcut (which we should not support, IMHO)

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+        output_path,
+        os.path.splitext(os.path.basename(filename_ck))[0] + '.xml'
+        )
+    et.write(filename_out, pretty_print=True, encoding='utf-8', xml_declaration=True)
+
+    return filename_out
+
+
+if __name__ == '__main__':
+    parser = ArgumentParser(description='Convert ReSpecTh XML file to ChemKED '
```
bryanwweber: In cases like this, I prefer moving the argument to the next line so that the whole string fits on a single line. I think its a little easier to read, YMMV

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+    return datapoints
+
+
+def convert_ReSpecTh(filename_xml, output='', file_author='', file_author_orcid=''):
+    """Convert ReSpecTh XML file to ChemKED YAML file.
+
+    Args:
+        filename_xml (`str`): Name of ReSpecTh XML file to be converted.
+        output (`str`, optional): Output path for converted file.
```
bryanwweber: This is a folder, right? IMHO, path is ambiguous if you mean folder.

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+if __name__ == '__main__':
+    parser = ArgumentParser(description='Convert ReSpecTh XML file to ChemKED '
+                                        'YAML file.'
+                            )
+    parser.add_argument('-i', '--input',
+                        type=str,
+                        required=True,
+                        help='Input XML filename'
+                        )
+    parser.add_argument('-o', '--output',
```
bryanwweber: Not sure this argument is more useful than one that allows the user to set the filename (which could include a folder, if they do something like `'folder/filename.yaml'`)

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    parser.add_argument('-fo', '--file-author-orcid',
+                        dest='file_author_orcid',
+                        type=str,
+                        required=False,
+                        default='',
+                        help='File author ORCID'
+                        )
+
+    args = parser.parse_args()
+    convert_ReSpecTh(args.input, args.output,
```
bryanwweber: `convert_ReSpecTh` and `convert_to_ReSpecTh` are confusing as names for the functions :smiley: Can we be consistent? Should `convert_to_ReSpecTh` be a method of the `ChemKED` class? I'm not sure how a user would call it, at the moment, since this file is intended to be a CLI script, right? I guess they could import it and call it, but that seems confusing.

pyked/tests/test_chemked.py
```diff
@@ -17,6 +17,30 @@
 
 warnings.simplefilter('always')
 
+# Create temporary directory for tests that need to create files
+# Taken from http://stackoverflow.com/a/22726782/1569494
+try:
```
bryanwweber: In what situation will `tempfile` not be available? We only support Python 3.5 and up

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    filename_out = os.path.join(
+        output_path,
+        os.path.splitext(os.path.basename(filename_ck))[0] + '.xml'
+        )
+    et.write(filename_out, pretty_print=True, encoding='utf-8', xml_declaration=True)
+
+    return filename_out
+
+
+if __name__ == '__main__':
```
bryanwweber: Can you add this to a function called `main` or something, so that we can create entry points? See [ck2cti](https://github.com/Cantera/cantera/blob/master/interfaces/cython/cantera/ck2cti.py#L2297) for an example

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+    # Try to get reference info via DOI
+    try:
+        reference['doi'] = elem.attrib['doi']
+        ref = None
+        try:
+            ref = habanero.Crossref().works(ids=reference['doi'])['message']
+        except (HTTPError, habanero.RequestError):
+            print('DOI not found')
+            raise KeyError
```
kyleniemeyer: this `KeyError` is raised so that the outer `except` catches it, and switches to parsing the `preferredKey` attribute for reference information. `doi` isn't a required element of ReSpecTh files, so it shouldn't be a proper error—only if both `doi` *and* `preferredKey` are missing.

pyked/converters.py
kyleniemeyer: Similar to my response above, this `KeyError` trips the outer `except` to switch to parsing via `preferredKey`

pyked/converters.py
kyleniemeyer: No, these aren't provided in ReSpecTh files.

pyked/converters.py
kyleniemeyer: Does using `get` simplify this at all? It would still be a `get`, then `if ... != None`, right?

pyked/converters.py
kyleniemeyer: There are only two conversions, for pressure and temperature, and they are already in an `if-elif` block:
```
    if ign_target == 'P':
        ign_target = 'pressure'
    elif ign_target == 'T':
        ign_target = 'temperature'
```
unless I'm missing what you mean?

pyked/converters.py
kyleniemeyer: It is `None`, I believe, so I'll change it

pyked/converters.py
kyleniemeyer: Not really sure what you mean here...

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: Actually this is an empty list, so I'm switching back to just `if not dataGroups`

pyked/converters.py
bryanwweber: I mean do something like `temp_prop = prop.attrib['name'].replace(' ', '-')` then check the value of `temp_prop`, and if it passes, then fill the value into `property_id`. What happens now is the `property_id` gets a value put into it, then the value is checked, and it errors out if the value doesn't pass. However, I think (might be wrong) there could be a case where the check is skipped for some reason (perhaps in the future if we add some other code or something, or someone catches that error upstream), and it could end up with invalid values in the `property_id` dictionary. Maybe this isn't a real concern, so if you think I'm off base, just ignore me 😄

pyked/converters.py
bryanwweber: I mean, combine the ones for pressure and temperature with the ones for CH* and OH*. Is there a reason they're split into two blocks?

pyked/converters.py
bryanwweber: > this `KeyError` is raised so that the outer except catches it

Right, I'm suggesting to use one of our custom errors, in case someone has intercepted `KeyError`s somewhere upstream. Although, given your second comment, its maybe better to just raise the appropriate error here, rather than relying on the outer `except` to catch and re-raise? There's probably also a way to simplify this logic that doesn't rely on nested try-except statements (I was pretty confused the first time I read through this).

pyked/converters.py
bryanwweber: ☹️

pyked/converters.py
bryanwweber: IMHO, using `get` or `if key in dict:` is easier to read, clearer (especially `if key in dict`) and more "pythonic", whatever that means 😸  FWIW, there's also some evidence (https://stackoverflow.com/a/20308657/2449192) that try/except is slower.

pyked/chemked.py
```diff
@@ -233,6 +239,30 @@ def get_dataframe(self, output_columns=None):
...
+    def write_file(self, filename, *, overwrite=False):
+        """Write new ChemKED YAML file based on object.
+
+        Arguments:
+            filename (`str`): Filename for target YAML file
+            overwrite (`bool`, optional): Whether to overwrite file with given name if present.
+                Must be supplied as a keyword-argument.
+
+        Raises:
+            `NameError`: If `filename` is already present, and `overwrite` is not `True`.
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+import os
+from argparse import ArgumentParser
+from warnings import warn
+import numpy
+
+from requests.exceptions import HTTPError, ConnectionError
+import habanero
+import pint
+
+try:
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    Returns:
+        properties (`dict`): Dictionary with file metadata
+    """
+    properties = {}
+
+    properties['file-author'] = {'name': ''}
+    try:
+        properties['file-author']['name'] = root.find('fileAuthor').text
+    except AttributeError:
+        raise AttributeError('Error: no fileAuthor given')
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+
+    # Default version is 0
+    version = '0'
+    elem = root.find('fileVersion')
+    if elem is None:
+        print('Warning: no fileVersion given')
+    else:
+        try:
+            version = elem.find('major').text + '.' + elem.find('minor').text
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,741 @@
...
+    Args:
+        root (`etree.Element`): Root of ReSpecTh XML file
+
+    Returns:
+        properties (`dict`): Dictionary with reference information
+    """
+    reference = {}
+    elem = root.find('bibliographyLink')
+    if elem is None:
+        raise AttributeError('Error: missing bibliographyLink')
```
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: Alright, I think I've improved this stuff.

pyked/converters.py
kyleniemeyer: Hopefully this is better now too.

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: Irritatingly, `etree.Element`s are not *actually* dictionaries, and so `if key in elem` doesn't work. However, `get` does work.

Though it looks like I could do `attributes = elem.attrib` to get a true dict, and then proceed.

kyleniemeyer: (Yeah, that looks a lot better now)

pyked/converters.py
kyleniemeyer: 👍  I see what you mean now

pyked/converters.py
kyleniemeyer: OK- not sure if we would actually face a problem here, but easy enough to do this.

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: Sure, seems reasonable.

pyked/converters.py
kyleniemeyer: how about `output_path (`str`, optional): Output directory path for converted file.` ?

pyked/converters.py
kyleniemeyer: yeah, probably not actually necessary.

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: Yeah, this isn't necessary anyways, I think—I believe I intended on coming back to this, but I'll just remove it altogether.

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: :

pyked/converters.py
kyleniemeyer: It turns out `compression-time` isn't supported by ReSpecTh anyways, so I'll remove all this.

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: Sure, this is just a convenience thing to generate a `common-properties` block like most other files. This is easy enough, so I will leave it.

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: OK, can do.

pyked/converters.py
kyleniemeyer: 👍

pyked/converters.py
kyleniemeyer: ok, can just do an output filename option.

pyked/converters.py
kyleniemeyer: yeah, i think this is more useful as a CLI script, and I think it makes sense to leave the two converters together. I can call them `convert_from_ReSpecTh` and `convert_to_ReSpecTh` if that helps

pyked/tests/test_chemked.py
kyleniemeyer: Ah, the glories of only supporting recent versions...

pyked/converters.py
bryanwweber: If it is going to be joined with the `filename_ck`, then I think it should be `output_folder`. But I don't know that we need this argument at all, since I think if people want a folder then can include it in `filename_ck` (but maybe need to check that). Is there a reason you want a folder argument like this?

pyked/converters.py
bryanwweber: OK, can you make sure to remove the comment so we don't keep coming back here? It looks like a TODO :smile:

pyked/converters.py
bryanwweber: I think we can do something like (in chemked.py) `from .converters import covert_to_ReSpecTh` and then in the `ChemKED` class call that function, provided that it can take a dictionary input, and there are no problems with circular dependencies. Maybe something to consider for the future. Regardless, there needs to be a function in this module to call the `to_ReSpecTch` converter so we can set up an entry point.

pyked/converters.py
bryanwweber: And also see the discussion on the comment about the wording of this option in the `convert_from_reSpecTh`

pyked/converters.py
bryanwweber: Also, what about `ChemKED_to_ReSpecTh` and `ReSpecTh_to_ChemKED` for function names? The it is consistent and obvious what each does (when we add PrIMe, is `convert_to_ReSpecTh` converting PrIMe or ChemKED?)

pyked/converters.py
kyleniemeyer: I just got rid of the path-related argument altogether.

pyked/converters.py
kyleniemeyer: ok, this naming suggestion sounds good. I can also add the ability to handle a dict input to `ChemKED_to_ReSpecTh`

kyleniemeyer:
Wahoo! Coverage back to 100%

kyleniemeyer:
And, with a `ChemKED` class method to convert to ReSpecTh, that is also available via the command line script

kyleniemeyer:
OK, now this should have everything: 100% coverage, `CHANGELOG` updated, and added a brief example on conversation to the docs.

pyked/tests/test_chemked.py
```diff
@@ -177,6 +180,185 @@ def test_many_species(self, pd):
...
+        filename = pkg_resources.resource_filename(__name__, file_path)
+        c = ChemKED(filename)
+
+        c.experiment_type = experiment_type
+
+        with pytest.raises(NotImplementedError) as excinfo:
+            c.convert_to_ReSpecTh('test.xml')
+        assert 'Only ignition delay type supported for conversion.' in str(excinfo.value)
+
+    def test_conversion_datapoints_composition_missing_inchi(self):
```
bryanwweber: There's no assertion in this test

pyked/tests/test_chemked.py
```diff
@@ -177,6 +180,185 @@ def test_many_species(self, pd):
...
+                                              'species-name': 'O2'},
+                                             {'amount': Q_(0.8, 'dimensionless'),
+                                              'species-name': 'Ar'}
+                                             ]
+
+        with TemporaryDirectory() as temp_dir:
+            newfile = os.path.join(temp_dir, 'test.xml')
+            c.convert_to_ReSpecTh(newfile)
+
+    def test_conversion_datapoints_different_composition(self):
```
bryanwweber: There's no assertion in this test

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+                prop.set('description', '')
+                prop.set('name', prop_name)
+                prop.set('units', str(getattr(self.datapoints[0], attribute).units))
+                idx = 'x{}'.format(len(property_idx) + 1)
+                property_idx[idx] = prop_name
+                prop.set('id', idx)
+                prop.set('label', labels[prop_name])
+
+        if 'composition' not in common:
+            for species in self.datapoints[0].composition:
```
bryanwweber: The composition of all the datapoints are not guaranteed to have the same species, so this has to be changed. You have a test for this, but it doesn't assert anything, so missed this case

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+        datagroup_link = etree.SubElement(datagroup, 'dataGroupLink')
+        datagroup_link.set('dataGroupID', '')
+        datagroup_link.set('dataPointID', '')
+
+        property_idx = {}
+        labels = {'temperature': 'T', 'pressure': 'P',
+                  'ignition delay': 'tau', 'pressure rise': 'dP/dt',
+                  }
+
+        for prop_name in datagroup_properties:
```
bryanwweber: I tried to simplify the next 40 lines or so, and also account for the fact that different datapoints can have different units for the quantity, so we have to convert, we can't assume all the units are the same as the first datapoint (if that datapoint even has the property we're interested in). The following can replace line 377 to line 419 (but doesn't account for my comment below on line 392):

```
        for prop_name in datagroup_properties:
            attribute = prop_name.replace(' ', '_')
            prop_indices = [i for i, dp in enumerate(self.datapoints)
                            if getattr(dp, attribute) is not None]
            if prop_name in common or not prop_indices:
                continue

            prop = etree.SubElement(datagroup, 'property')
            prop.set('description', '')
            prop.set('name', prop_name)
            units = str(getattr(self.datapoints[prop_indices[0]], attribute).units)
            prop.set('units', units)
            idx = 'x{}'.format(len(property_idx) + 1)
            property_idx[idx] = {'name': prop_name, 'units': units}
            prop.set('id', idx)
            prop.set('label', labels[prop_name])

        if 'composition' not in common:
            for species in self.datapoints[0].composition:
                prop = etree.SubElement(datagroup, 'property')
                prop.set('description', '')

                idx = 'x{}'.format(len(property_idx) + 1)
                property_idx[idx] = {'name': species['species-name']}
                prop.set('id', idx)
                prop.set('label', '[' + species['species-name'] + ']')
                prop.set('name', 'composition')
                prop.set('units', self.datapoints[0].composition_type)

                species_link = etree.SubElement(prop, 'speciesLink')
                species_link.set('preferredKey', species['species-name'])
                if species.get('InChI'):
                    species_link.set('InChI', species['InChI'])

        for dp in self.datapoints:
            datapoint = etree.SubElement(datagroup, 'dataPoint')
            for idx, val in property_idx.items():
                value = etree.SubElement(datapoint, idx)
                # handle regular properties a bit differently than composition
                if val['name'] in datagroup_properties:
                    quantity = getattr(dp, val['name'].replace(' ', '_')).to(val['units'])
                    value.text = str(quantity.magnitude)
                else:
                    # composition
                    for item in dp.composition:
                        if item['species-name'] == val['name']:
                            value.text = str(item['amount'].magnitude)
```

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+                            value.text = str(item['amount'].magnitude)
+
+        # if RCM and has volume history, need a second dataGroup
+        if (len(self.datapoints) > 1 and
+            any([getattr(dp, 'volume_history', None) for dp in self.datapoints])
+            ):
+            raise NotImplementedError('Error: ReSpecTh files do not support multiple datapoints '
+                                      'with a volume history.'
+                                      )
+            # TODO: what if they share the same history? Does this happen?
```
bryanwweber: This should not happen, unless the experimentalist is taking shortcuts, so you can delete this comment

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+                if property_idx[idx] in datagroup_properties:
+                    value.text = str(getattr(dp, property_idx[idx].replace(' ', '_')).magnitude)
+                else:
+                    # composition
+                    for item in dp.composition:
+                        if item['species-name'] == property_idx[idx]:
+                            value.text = str(item['amount'].magnitude)
+
+        # if RCM and has volume history, need a second dataGroup
+        if (len(self.datapoints) > 1 and
```
bryanwweber: Consider the following for clarity:

```
        has_volume_history = any([hasattr(dp, 'volume_history') for dp in self.datapoints])
        if len(self.datapoints) > 1 and has_volume_history:
```

At the least, I would say to use `hasattr` rather than `getattr(..., None)`.

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+            any([getattr(dp, 'volume_history', None) for dp in self.datapoints])
+            ):
+            raise NotImplementedError('Error: ReSpecTh files do not support multiple datapoints '
+                                      'with a volume history.'
+                                      )
+            # TODO: what if they share the same history? Does this happen?
+        elif getattr(self.datapoints[0], 'volume_history', None):
```
bryanwweber: Again, consider `hasattr` for clarity. Also, should it be its own `if` rather than `elif`? This isn't really connected to the previous `if`

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+            datagroup = etree.SubElement(root, 'dataGroup')
+            datagroup.set('id', 'dg1')
```
bryanwweber: Should be `'dg2'`?

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+            prop.set('label', 'V')
+
+            for time, volume in zip(volume_history.time, volume_history.volume):
+                datapoint = etree.SubElement(datagroup, 'dataPoint')
+                value = etree.SubElement(datapoint, time_idx)
+                value.text = str(time.magnitude)
+                value = etree.SubElement(datapoint, volume_idx)
+                value.text = str(volume.magnitude)
+
+        # In ReSpecTh files all datapoints share ignition type
```
bryanwweber: What if the ChemKED file in question has multiple ignition types?

docs/ck-tutorial.rst
```diff
@@ -490,6 +490,52 @@ shock tube at Stanford.
...
+PyKED provides converter functions from (and to) the ReSpecTh XML format
+[Varga2015a]_, [Varga2015b]_, though ChemKED files created in this manner may
+require manual edits to fully satisfy our schema.
+
+Given a ReSpecTh XML file ``file.xml``, a corresponding ChemKED file can be created
+either from the command line
+
+.. code-block:: bash
+
+    python converter -i file.xml -o file.yaml
```
kyleniemeyer: OK, I thought that by defining the entry point, you don't need the module: 
```
 entry_points = {
        'console_scripts': ['converter=pyked.converters:main'],
    }
```
But, I can make the title more specific—does `respth2ck` imply a direction?

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
+"""Module with converters from other formats.
+"""
+
+# Standard libraries
+import os
+from argparse import ArgumentParser
+from warnings import warn
+import xml.etree.ElementTree as etree
+
+import numpy
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+import pint
+
+# Local imports
+from .validation import yaml, property_units
+from .utils import units as unit_registry
+from ._version import __version__
+
+
+# Valid properties for ReSpecTh dataGroup
+datagroup_properties = ['temperature', 'pressure', 'ignition delay', 'pressure rise']
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+
+# Exceptions
+class ParseError(Exception):
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+    """Base class for errors."""
+    pass
+
+class KeywordError(ParseError):
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+    """Raised for errors in keyword parsing."""
+
+    def __init__(self, *keywords):
+        self.keywords = keywords
+
+    def __str__(self):
+        return repr('Error: {}.'.format(self.keywords[0]))
+
+class MissingElementError(KeywordError):
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+    """Raised for missing required elements."""
+
+    def __str__(self):
+        return repr('Error: required element {} is missing.'.format(
+            self.keywords[0]))
+
+class MissingAttributeError(KeywordError):
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+
+    Args:
+        root (`etree.Element`): Root of ReSpecTh XML file
+
+    Returns:
+        properties (`dict`): Dictionary with file metadata
+    """
+    properties = {}
+
+    properties['file-author'] = {'name': ''}
```
kyleniemeyer: nice 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+        root (`etree.Element`): Root of ReSpecTh XML file
+
+    Returns:
+        properties (`dict`): Dictionary with experiment type and apparatus information.
+    """
+    properties = {}
+    if root.find('experimentType').text == 'Ignition delay measurement':
+        properties['experiment-type'] = 'ignition delay'
+    else:
+        #TODO: support additional experimentTypes
```
kyleniemeyer: I suppose that adding additional experiments would require a major overhaul of this module, so keeping track of this one place isn't that important...

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+        raise NotImplementedError(root.find('experimentType').text + ' not (yet) supported')
+
+    properties['apparatus'] = {'kind': '', 'institution': '', 'facility': ''}
+    try:
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+                    warn('Missing InChI for species ' + spec['species-name'])
+                    pass
+
+                # amount of that species
+                spec['amount'] = [float(child.find('amount').text)]
+
+                properties['composition']['species'].append(spec)
+
+                # check consistency of composition type
+                if not composition_type:
```
kyleniemeyer: Seems like both work, but from what I can tell yes, when explicitly testing for `None`, `is None` is more appropriate.

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+                    composition_type = child.find('amount').attrib['units']
+                elif composition_type != child.find('amount').attrib['units']:
+                    raise KeywordError('inconsistent initial composition units')
```
kyleniemeyer: is that necessary? A more informative message might just be `'Initial composition units need to be consistent'`

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+            assert composition_type in ['mole fraction', 'mass fraction']
```
kyleniemeyer: actually, based on the ReSpecTh documentation, the valid units for `composition` are `mole fraction, percent, ppm, ppb`. I assume `percent` means `mole percent`, though I'm not sure... but it looks like they don't actually support `mass fraction`

kyleniemeyer: We can at least support conversion of all the three that ChemKED supports, and modify if we encounter ReSpecTh files defined differently

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+
+
+def get_ignition_type(root):
+    """Gets ignition type and target.
+
+    Args:
+        root (`etree.Element`): Root of ReSpecTh XML file
+
+    Returns:
+        properties (`dict`): Dictionary with ignition type/target information
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+def ReSpecTh_to_ChemKED(filename_xml, filename_ck='', file_author='', file_author_orcid=''):
+    """Convert ReSpecTh XML file to ChemKED YAML file.
+
+    Args:
+        filename_xml (`str`): Name of ReSpecTh XML file to be converted.
+        filename_ck (`str`, optional): Name of output ChemKED file to be produced.
+        file_author (`str`, optional): Name to override original file author
+        file_author_orcid (`str`, optional): ORCID of file author
+    """
+    from pyked.chemked import ChemKED
```
kyleniemeyer: If this is a module-level import, then there will be a circular import when `chemked.py` tries to do `from .converters import datagroup_properties`...

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+
+    assert os.path.isfile(filename_xml), 'Error: ' + filename_xml + ' file missing'
```
kyleniemeyer: Yeah, this is unnecessary

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+    properties['common-properties'] = get_common_properties(root)
+
+    # Determine definition of ignition delay
+    properties['common-properties']['ignition-type'] = get_ignition_type(root)
+
+    # Now parse ignition delay datapoints
+    properties['datapoints'] = get_datapoints(root)
+
+    # Ensure inclusion of pressure rise or volume history matches apparatus.
+    if ('pressure-rise' in properties['common-properties'] or
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+        any([dp for dp in properties['datapoints'] if dp.get('pressure-rise')])
+        ) and properties['apparatus']['kind'] == 'rapid compression machine':
+        raise KeywordError('Pressure rise cannot be defined for RCM.')
+
+    if ('volume-history' in properties['common-properties'] or
+        any([dp for dp in properties['datapoints'] if dp.get('volume-history')])
+        ) and properties['apparatus']['kind'] == 'shock tube':
+        raise KeywordError('Volume history cannot be defined for shock tube.')
+
+    # apply any overrides
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+
+    # compression time doesn't belong in common-properties
+    properties['common-properties'].pop('compression-time', None)
+
+    # set output filename and path
+    if not filename_ck:
+        filename_ck = os.path.splitext(os.path.basename(filename_xml))[0] + '.yaml'
+
+    with open(filename_ck, 'w') as outfile:
+        outfile.write(yaml.dump(properties, default_flow_style=False))
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+    print('Converted to ' + filename_ck)
+
+    # now validate
+    ChemKED(yaml_file=filename_ck)
+
+
+def main(argv):
+    """
+    """
+    from pyked.chemked import ChemKED
```
kyleniemeyer: Same issue as before—a module-level import would lead to circular import, I think

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+                        dest='file_author_orcid',
+                        type=str,
+                        required=False,
+                        default='',
+                        help='File author ORCID'
+                        )
+
+    args = parser.parse_args(argv)
+
+    if (os.path.splitext(args.input)[1] == '.xml' and
```
kyleniemeyer: 👍

pyked/converters.py
```diff
@@ -0,0 +1,522 @@
...
+          os.path.splitext(args.output)[1] == '.yaml'
+          ):
+        raise KeywordError('Cannot convert .yaml to .yaml')
+
+    else:
+        raise KeywordError('Input/output args need to be .xml/.yaml')
+
+
+if __name__ == '__main__':
+    main(sys.argv[1:])
```
kyleniemeyer: whoops...

test-environment.yaml
```diff
@@ -14,3 +14,4 @@ dependencies:
   - uncertainties
   - habanero
   - orcid
+  - lxml>=3.8.0
```
kyleniemeyer: 👍

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+        Raises:
+            `NameError`: If ``filename`` is already present, and ``overwrite`` is not ``True``.
+
+        Example:
+            >>> dataset = ChemKED(yaml_file)
+            >>> dataset.write_file(new_yaml_file)
+        """
+        # Ensure file isn't already present
+        if exists(filename) and not overwrite:
+            raise NameError(filename + ' already present. Specify "overwrite=True" '
```
kyleniemeyer: 👍

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+        apparatus = etree.SubElement(root, 'apparatus')
+        kind = etree.SubElement(apparatus, 'kind')
+        kind.text = self.apparatus.kind
+
+        common_properties = etree.SubElement(root, 'commonProperties')
+        # ChemKED objects have no common properties once loaded. Check for properties
+        # among datapoints that tend to be common
+        common = []
+        composition = self.datapoints[0].composition
+        # Composition type *has* to be the same
```
kyleniemeyer: This particular location is testing whether all of the datapoints have the same composition, so I think it's appropriate here (if the others don't match datapoint 0, then it isn't a common property- good!)

However, on line 391, where the list of species are being added as `property` elements, I *do* make the assumption that the all datapoints at least include the same set of species, which isn't correct.

kyleniemeyer: ReSpecTh doesn't support datapoints with different composition types, though...

kyleniemeyer: Alright, I think I figured out how to handle this.

kyleniemeyer: nope, harder than I thought.

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+            # initial composition is common
+            common.append('composition')
+            prop = etree.SubElement(common_properties, 'property')
+            prop.set('name', 'initial composition')
+
+            for species in composition:
+                component = etree.SubElement(prop, 'component')
+                species_link = etree.SubElement(component, 'speciesLink')
+                species_link.set('preferredKey', species['species-name'])
+                if species.get('InChI'):
```
kyleniemeyer: 👍

pyked/chemked.py
kyleniemeyer: In ReSpecTh, different datapoints *can't* have different units—the units are defined in a `property` element, and the values are given as numbers only in `dataPoint` elements.

pyked/chemked.py
kyleniemeyer: 👍

pyked/chemked.py
kyleniemeyer: 👍

pyked/chemked.py
kyleniemeyer: 👍

pyked/chemked.py
kyleniemeyer: yep, not that this is used, from what I can tell...

pyked/chemked.py
kyleniemeyer: ¯\_(ツ)_/¯

Can't be converted... guess it should be an error.

pyked/chemked.py
bryanwweber: Right, I think that's what this does. In CK, the different data points *can* have different units, so we have to make sure to use Pint to convert them to a consistent set, which is what this does.

docs/ck-tutorial.rst
bryanwweber: By defining the entry point, you don't need to invoke Python at all (and I believe what you have in the docs will cause an error because it can't find a script called converter). The entry point is a bash/batch script. You can also invoke a particular module (I think if it has the `__main__` function defined?) by doing `python -m module.name args`, but that's separate from the entry point.

I think `respth2ck` implies XML to YAML, but as I said above, I think its worth it to have 2 explicit scripts and one combined script. Maybe `respth2ck`, `ck2respth`, and `convert_ck`?

pyked/converters.py
bryanwweber: Right, I think `None` is falsey, but `''` is also falsey, and is a (hypothetically) possible value for `composition_type`

pyked/converters.py
bryanwweber: I'm thinking that a user would like to see which units are inconsistent, to help with troubleshooting.

pyked/converters.py
bryanwweber: `ppm` and `ppb` are (usually) just mole fractions with other scaling. However, they could also be mass fractions I suppose, as could percent... is there anything in the ReSpecTh spec about this?

pyked/chemked.py
bryanwweber: > This particular location is testing whether all of the datapoints have the same composition, so I think it's appropriate here (if the others don't match datapoint 0, then it isn't a common property- good!)

Is this checking that the `composition_type`s are all the same? If not, I think it should because, hypothetically anyways, the species and numbers could all be equal (i.e., the `composition`), but some could be mass fraction and some could be mole fraction (i.e., the `composition_type`s could be different). This isn't likely to happen, but I think we should code more defensively because you never know what those crazy experimentalists are going to try to do next...

In addition, the comment isn't correct for a CK instance... not sure if you meant it to refer to the CK class or the ReSpecTh file. Maybe that has to be one of the cases we can't convert to ReSpecTh, if there are multiple `composition_type`s in a set of datapoints.

pyked/converters.py
bryanwweber: You can do `from . import chemked` and then below use `chemked.ChemKED(...`. That doesn't raise any circular import errors, I don't think

pyked/chemked.py
bryanwweber: Does the :+1: mean for just the `hasattr`, or also for the new `if` instead of `elif`?

pyked/converters.py
kyleniemeyer: Hah, that's a joke... what I copied there is literally all it specifies.

pyked/converters.py
kyleniemeyer: OK, can do.

pyked/converters.py
kyleniemeyer: OK, trying this.

pyked/chemked.py
kyleniemeyer: OK, now we raise an error if datapoints have multiple composition types, because ReSpecTh doesn't support this (from what I can tell).

pyked/chemked.py
```diff
@@ -233,6 +241,238 @@ def get_dataframe(self, output_columns=None):
...
+                amount.set('units', composition_type)
+                amount.text = str(species['amount'].magnitude)
+
+        # If multiple datapoints present, then find any common properties. If only
+        # one datapoint, then composition should be the only "common" property.
+        if len(self.datapoints) > 1:
+            for prop_name in datagroup_properties:
+                attribute = prop_name.replace(' ', '_')
+                quantity = getattr(self.datapoints[0], attribute)
+                if (quantity != None and
```
kyleniemeyer: nice 👍

pyked/chemked.py
kyleniemeyer: gotcha. I updated the middle block with my other change to handle (potential) variable species sets, but otherwise looks good.

pyked/chemked.py
kyleniemeyer: Yeah, got this now i think.

kyleniemeyer:
🙌 

(This one took a while...)

PR #75, opened 2017-09-25
Appveyor cov

bryanwweber:
 - Upload coverage from Appveyor to codecov
 - Only run 1 Travis job per OS to speed things up (hopefully)

@pr-omethe-us/chemked


bryanwweber:
So it saved about 30 seconds on Linux to combine the Travis runs. I think the main benefit is that only one macOS job has to run, which means less waiting in the queue (hopefully), but as we add more tests, the benefit is going to drop... Thoughts?

kyleniemeyer:
I guess it doesn't seem that big of a deal, but streamlines things a little bit

bryanwweber:
The test failures here will be resolved when #77 is merged

PR #80, opened 2017-10-02
Add tests for composition type names

bryanwweber:
- [x] Tests added

 - Add test to check that the composition type is stored properly in the DataPoint

@pr-omethe-us/chemked


bryanwweber:
OK I finally fixed the failure here. Not sure why conda itself wasn't being updated during the `update --all` stage, and also not sure why conda has to be updated to install the most recent versions of the packages since it was only a patch version change, but here we are...

PR #81, opened 2017-10-04
made test_incorrect_doi_period_at_end test match description

kyleniemeyer:
- [X] Tests fixed: `test_incorrect_doi_period_at_end`

 - Makes test `test_incorrect_doi_period_at_end` match description

@pr-omethe-us/chemked

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Tests that the composition type is stored properly in the `DataPoint`
 
 ### Fixed
+- Fixes `test_incorrect_doi_period_at_end` to actually test DOI with period
 
 ### Changed
 - Conda builds are now noarch - one package for all Pythons!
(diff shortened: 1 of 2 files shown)


bryanwweber:
Do we need this test? What is it testing for?

kyleniemeyer:
If I recall, some of the ReSpecTh files I saw had periods at the end, and some didn't—so this was just a little test to make sure we can handle that.

bryanwweber:
But a period at the end of a DOI is a different DOI, no? So if they have a period at the end, its automatically an invalid DOI regardless of why it has a period at the end, unless that period is meant to be stripped off in the processing?

kyleniemeyer:
Ohh I just figured out/remembered what this test was for... it wasn't about a period at the end of the DOI, but at the end of the reference. 

In the case of an invalid reference, the converter puts whatever is given in the `detail` field in our `reference` (with a message so the user can manually correct). The period issue is minor, but a sentence is added after the reference data is copied, so the code checks for a period and adds if there isn't one. This was testing the other branch of that if statement, I think.

bryanwweber:
OK that makes more sense. I rebased so its just one commit to avoid the noise and churn. Merged!

Issue #54, opened 2017-05-22
Composition in datapoint should be a namedtuple


bryanwweber:
As the title suggests, this would make it easier to document the fields of the composition

PR #101, opened 2017-11-29
[WIP] Changing composition to namedtuple

EthanBesse:
Related to #54 

I have changed `composition` on line 613in chemked.py to a namedtuple as well as began converting code in the loop starting on line 623. Could I have this looked over for correctness?

### DIFF
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -37,6 +37,8 @@ Apparatus.kind.__doc__ = '(`str`) The kind of experimental apparatus'
 Apparatus.institution.__doc__ = '(`str`) The institution where the experiment is located'
 Apparatus.facility.__doc__ = '(`str`) The particular experimental facility at the location'
 
+Composition = namedtuple('Composition', 'species_name InChI SMILES atomic_composition amount')
+
 
 class ChemKED(object):
     """Main ChemKED class.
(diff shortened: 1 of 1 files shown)


pyked/chemked.py
```diff
@@ -37,6 +37,8 @@
 Apparatus.institution.__doc__ = '(`str`) The institution where the experiment is located'
 Apparatus.facility.__doc__ = '(`str`) The particular experimental facility at the location'
 
+Composition = namedtuple('Composition', 'species_name InChI SMILES atomic_composition amount')
```
bryanwweber: This looks fine, make sure you add the documentation as is done for the other namedtuples

pyked/chemked.py
```diff
@@ -608,12 +610,20 @@ def __init__(self, properties):
...
-
+        # composition = deepcopy(properties['composition']['species'])
+        composition = []
+        for spec in properties['composition']['species']:
+            species_name = spec['species-name']
+            amount = spec['amount']
+            InChI = spec.get('InChI')
+            SMILES = spec.get('SMILES')
+            atomic_composition = spec.get('atomic-composition')
+            composition.append(Composition(species_name=species_name, InChI=InChI, SMILES=SMILES, atomic_composition=atomic_composition, amount=amount))
```
bryanwweber: Try to make sure the line length is less than 100 characters (I can't tell if this is or not)

pyked/chemked.py
```diff
@@ -608,12 +610,20 @@ def __init__(self, properties):
...
+        print(composition)
```
bryanwweber: Make sure to delete this

pyked/chemked.py
```diff
@@ -608,12 +610,20 @@ def __init__(self, properties):
...
         for idx, species in enumerate(composition):
```
bryanwweber: You can combine this loop with the previous loop by doing the processing on `spec['amount']` when you get the value out of the dictionary. That will be more efficient. You can probably just delete this line, and move your line 621 (`composition.append(...)`) to line 659. You'll also have to change lines 624->626 to use `spec['amount']` rather than `species['amount']` but everything else should stay the same

pyked/chemked.py
```diff
@@ -646,11 +656,11 @@ def __init__(self, properties):
                 else:
                     raise ValueError('uncertainty-type must be one of "absolute" or "relative"')
 
-            composition[idx]['amount'] = quant
-        setattr(self, 'composition', composition)
+            #composition[idx]['amount'] = quant
 
-        self.equivalence_ratio = properties.get('equivalence-ratio')
+            setattr(self, 'composition', composition)
```
bryanwweber: This should only be set once, not on every time through the for loop (I think your indentation is a bit off)

bryanwweber:
@EthanBesse Sorry for the delay! I added some comments on the code. You might also want to make sure that your GitHub Desktop is set up properly, it looks like these commits were made with the "Kypore" email account. You can just go into the settings of GHD and in the Git tab, change the email address there to be the one associated with your EthanBesse account

PR #104, opened 2017-12-11
[WIP] Add time histories

### LINKED ISSUE #91: How can we include RCM parameters sensibly?
After discussions with Scott and Stephen, we have determined that we need to add several RCM experimental parameters to the ChemKED file as optional fields. We need to determine:

a) what those fields are
b) where they should go
c) how to make them machine readable

We also need to determine what fields belong in a general experimental database, and what belongs in an individual ChemKED file.

bryanwweber:
- [ ] Fixes #91 
- [ ] Tests added
- [ ] Documentation updated

- This PR adds a `time-histories` field to encapsulate all of the time-varying properties of an experiment, including pressure, temperature, volume, absorption, emission, and piston position

@pr-omethe-us/chemked


kyleniemeyer:
Was this expected to fail (if still in progress), or no?

bryanwweber:
It's a flake8 failure, so I'd say, yes because WIP, but not expected 😄

pyked/tests/testfile_rcm2.yaml
```diff
@@ -32,12 +32,21 @@ datapoints:
...
+        - uncertainty: 0.01
+          uncertainty-type: relative
+      compression-time:
+        - 38.0 ms
+      stroke:
+        - 10.0 inch
+      clearance:
+        - 2.5 cm
+      compression-ratio:
+        - '12.0'
```
kyleniemeyer: are the quote marks needed here?

pyked/validation.py
```diff
@@ -185,17 +190,60 @@ def _validate_isvalid_unit(self, isvalid_unit, field, value):
...
+            self._error(field, 'not enough columns in the values')
+
     def _validate_isvalid_quantity(self, isvalid_quantity, field, value):
         """Checks for valid given value and appropriate units.
 
         Args:
             isvalid_quantity (`bool`): flag from schema indicating quantity to be checked.
             field (`str`): property associated with quantity in question.
-            value (`str`): string of the value of the quantity
+            value (`list`): list with the first element a string of value with units
```
kyleniemeyer: This doesn't read correctly to me: `list with the first element a string of value with units`

bryanwweber: Oh yeah, that's really bad grammar hah

pyked/tests/testfile_rcm2.yaml
bryanwweber: Ah, no, shouldn't be, I was testing a weird failure mode. Thanks for the catch

bryanwweber:
Once the tests pass, this should be good to merge :+1:

PR #109, opened 2018-03-07
Fix ORCID API calls

bryanwweber:
- [x] Documentation updated

- Replace the python-orcid package because it doesn't support the new ORCID API

@pr-omethe-us/chemked

### DIFF
--- a/build-environment.yaml
+++ b/build-environment.yaml
@@ -10,7 +10,6 @@ dependencies:
   - pandas
   - uncertainties
   - habanero>=0.6.0
-  - orcid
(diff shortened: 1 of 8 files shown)


kyleniemeyer:
Do we need some tests for the new function?

bryanwweber:
No, it should all be covered by the existing test suite

kyleniemeyer:
CHANGELOG?

kyleniemeyer:
I'm good with this other than that change.

bryanwweber:
In #110

PR #115, opened 2018-06-14
csv file time history

jsantner:
- [ ] Fixes #
- [x] Tests added
- [ ] Documentation updated

- Add test .yaml file that utilizes a .csv file for time-history
- Alter the validation module so that reading time-history from .csv does not create errors.

@pr-omethe-us/chemked

### DIFF
--- a/CITATION.md
+++ b/CITATION.md
@@ -11,7 +11,7 @@ A BibTeX entry for LaTeX users is
 ```TeX
 @misc{PyKED,
     author = {Kyle E Niemeyer and Bryan W Weber},
-    year = 2017,
+    year = 2018,
     title = {PyKED v0.4.1},
     doi = {10.5281/zenodo.597935},
     url = {https://github.com/pr-omethe-us/PyKED},
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -111,8 +111,10 @@ class ChemKED(object):
         if yaml_file is not None:
             with open(yaml_file, 'r') as f:
                 self._properties = yaml.safe_load(f)
+            directory = dirname(yaml_file)
         elif dict_input is not None:
             self._properties = dict_input
+            directory = None
         else:
             raise NameError("ChemKED needs either a YAML filename or dictionary as input.")
 
@@ -121,7 +123,7 @@ class ChemKED(object):
 
         self.datapoints = []
         for point in self._properties['datapoints']:
-            self.datapoints.append(DataPoint(point, dirname(yaml_file)))
+            self.datapoints.append(DataPoint(point, directory))
 
         self.reference = Reference(
             volume=self._properties['reference'].get('volume'),
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -2,7 +2,7 @@
 Main ChemKED module
 """
 # Standard libraries
-from os.path import exists
+from os.path import exists, isabs, dirname, join
 from collections import namedtuple
 from warnings import warn
 from copy import deepcopy
@@ -121,7 +121,7 @@ class ChemKED(object):
 
         self.datapoints = []
         for point in self._properties['datapoints']:
-            self.datapoints.append(DataPoint(point))
+            self.datapoints.append(DataPoint(point, dirname(yaml_file)))
 
         self.reference = Reference(
             volume=self._properties['reference'].get('volume'),
@@ -589,6 +589,7 @@ class DataPoint(object):
 
     Arguments:
         properties (`dict`): Dictionary adhering to the ChemKED format for ``datapoints``
+        directory (`str`, optional): Directory to look for auxiliary files
 
     Attributes:
         composition (`list`): List of dictionaries representing the species and their quantities
@@ -631,7 +632,7 @@ class DataPoint(object):
         'compression-ratio'
     ]
 
-    def __init__(self, properties):
+    def __init__(self, properties, directory=None):
         for prop in self.value_unit_props:
             if prop in properties:
                 quant = self.process_quantity(properties[prop])
@@ -685,7 +686,10 @@ class DataPoint(object):
                     values = np.array(hist['values'])
                 else:
                     # Load the values from a file
-                    values = np.genfromtxt(hist['values']['filename'], delimiter=',')
+                    filename = hist['values']['filename']
+                    if not isabs(filename):
+                        filename = join(directory, filename)
+                    values = np.genfromtxt(filename, delimiter=',')
 
                 time_history = TimeHistory(
                     time=Q_(values[:, time_col], time_units),
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -634,7 +634,7 @@ class DataPoint(object):
         'compression-ratio'
     ]
 
-    def __init__(self, properties, directory=None):
+    def __init__(self, properties, directory=''):
         for prop in self.value_unit_props:
             if prop in properties:
                 quant = self.process_quantity(properties[prop])
--- a/pyked/chemked.py
+++ b/pyked/chemked.py
@@ -2,13 +2,16 @@
 Main ChemKED module
 """
 # Standard libraries
-from os.path import exists, isabs, dirname, join
+from os.path import exists
 from collections import namedtuple
 from warnings import warn
 from copy import deepcopy
 import xml.etree.ElementTree as etree
 import xml.dom.minidom as minidom
 from itertools import chain
+from pathlib import Path
+from distutils.version import LooseVersion
+import sys
 
 import numpy as np
 
@@ -78,6 +81,13 @@ Composition.atomic_composition.__doc__ = '(`dict`) The atomic composition of the
 Composition.amount.__doc__ = '(`~pint.Quantity`) The amount of this species'
 
 
+if LooseVersion(sys.version) < '3.6':
+    # Allow old version of python to open Path objects.
+    oldopen = open
+    def open(file, mode='r', buffering=-1, encoding=None, errors=None, newline=None, closefd=True, opener=None):
+        return oldopen(str(file), mode, buffering, encoding, errors, newline, closefd, opener)
+
+
 class ChemKED(object):
     """Main ChemKED class.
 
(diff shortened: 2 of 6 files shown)


bryanwweber:
@jsantner Why are these changes needed?

jsantner:
@bryanwweber 
Previously, when loading a yaml file with a time history defined in a csv file, the validator gives this error message:

      File "c:\users\jsantne\documents\github\pyked\pyked\validation.py", line 255, in _validate_isvalid_history
        n_cols = len(value['values'][0])

    KeyError: 0

You can see this on the [Travis report](https://travis-ci.org/jsantner/PyKED/builds/392026211) after my second commit on this branch, where I had only added a test yaml file with time history defined in a csv file.

pyked/validation.py
```diff
@@ -252,14 +252,20 @@ def _validate_isvalid_history(self, isvalid_history, field, value):
...
-                       value['quantity']['column'],
-                       value.get('uncertainty', {}).get('column', 0)) + 1
-        if n_cols > max_cols:
-            self._error(field, 'too many columns in the values')
-        elif n_cols < max_cols:
-            self._error(field, 'not enough columns in the values')
+        # If reading from a file, the file will not be validated.
+        # A file can have an arbitrary number of columns, and the columns
+        # to be used are specified.
+        if type(value['values']) is list:
```
bryanwweber: Why not `'filename' not in value['values'].keys()`?

bryanwweber (changes requested):
@jsantner Thanks for submitting this! A few suggestions:

1. I think the assumption is that the CSV filename will be specified relative to the directory of the YAML file, hence the `directory` argument should be unnecessary. If that's not working, I'd be interested to see a failing test case.
2. If possible, I think we should load and check the CSV file as well. However, I think that will fit better when we refactor all of the validation, so we can skip it for now.

pyked/validation.py
jsantner: I agree that we can assume the CSV file is specified relative to the yaml file, but `DataPoint` doesn't know where the yaml file directory is unless it's specified as an argument to `__init__`, right? Is there a simpler way to deal with a relative path?

I tried using `'filename' not in value['values'].keys()` in a [previous commit](https://github.com/pr-omethe-us/PyKED/pull/115/commits/abe35c49c489450d088d4c0c9ee37dd1235daa21) but this fails when the values are given as a list. In that case, `value['values']` is a list, not a dict, so `if 'filename' not in value['values'].keys()` raises an error.

bryanwweber: As far as I know, just trying to open the file will try to open it relative to the working directory of the Python process. Are you saying that if I have a file structure like
```
|- database
|---butanol
|------file_1.yaml
|------file_1.csv
```

and I start Python in the `database` directory, and load `file_1.yaml` like
```
>>> ChemKED('butanol/file_1.yaml')
```
it won't work, because Python will assume the `file_1.csv` is relative to `database`, not `butanol`?

As on your other PR, I think the simpler/more "pythonic" way to do this is a `try...except` block, rather than checking the type of the value.

jsantner: Yes, that's exactly what I'm saying. I was using a script to read multiple yaml files in a complex directory structure within a `database` directory, and python was looking for the csv file in `database`, not in the folder with the yaml file.

I'm not sure that a `try...except` block would work well here. Are you thinking of something like? It seems more complex and confusing this way, and it puts a lot of code between the `try` and `except`
```
try:
    if 'filename' in value['values'].keys():
        # Don't do anything because csv files aren't validated
        pass
    else:
        # This should never happen. If vale['values'] is a dictionary with keys, 'filename' should be a key
        self._error(field, 'must include filename or list of values')
except KeyError:
        # value['values'] is probably a list.
        # Code from earlier that checks the number of columns
```

Since this PR is related to time histories, I have a somewhat related question for you that I just stumbled on. Let's say somebody has a csv file with three columns - time, pressure, and volume. Right now, these must be implemented as two separate `time-histories` and the csv will be loaded twice, right? Is there interest in allowing the user to specify multiple time-histories using a single file?

bryanwweber:
> Yes, that's exactly what I'm saying. I was using a script to read multiple yaml files in a complex directory structure within a database directory, and python was looking for the csv file in database, not in the folder with the yaml file.

OK, that's a case that we missed for sure. What if, rather than passing the directory name around, we turn the `yaml_file` into an instance of a [`Path`](https://docs.python.org/3/library/pathlib.html) object? Then we can find the path of the CSV file by doing `yaml_file.stem/csv_filename`. As of Python 3.6, the built-in open functions (which NumPy relies on) all accept instances of `Path` (see https://docs.python.org/3/library/functions.html#open), so we only really have to handle the Python 3.5 case, for which we can write a custom open function that just converts the filename to a string, something like

```
if sys.version < (3,6):
    oldopen = open
    def open(file, mode='r', buffering=-1, encoding=None, errors=None, newline=None, closefd=True, opener=None):
        return oldopen(str(file), mode, buffering, encoding, errors, newline, closefd, opener)
```

> Since this PR is related to time histories, I have a somewhat related question for you that I just stumbled on. Let's say somebody has a csv file with three columns - time, pressure, and volume. Right now, these must be implemented as two separate time-histories and the csv will be loaded twice, right? Is there interest in allowing the user to specify multiple time-histories using a single file?

The only problem I see with loading twice is that it might take some time to load the file from disk. I think it makes more sense to keep the specifications of the time histories separate, and try to cache the data in the csv file somehow, rather than loading it from disk twice.

pyked/validation.py
bryanwweber: Just to keep this focused on the code here, I moved the other discussion to the main comment thread. Anyhow, two things:
1) I want to validate csv files in the future, so we might as well set up for that here
2) The code could be

```
try:
    if 'filename' not in value['values'].keys():
        self._error(field, 'must include filename or list of values')
except TypeError:
    # Code from earlier that checks the number of columns
```
or
```
try:
    filename = value['values'].get('filename')
    if filename is None:
        self._error(field...
except AttributeError:
    # Code from earlier
```
which isn't all that confusing to me. The reason (to me) to avoid the `type` function is that it doesn't always handle inheritance in a straightforward way, so we'd be relying on the underlying YAML library to always return something that's a subclass of list. On the other hand, with the try-except, we're using the duck-typing in Python to try something that we expect to be the case, and catch the resulting errors.

pyked/chemked.py
```diff
@@ -685,7 +688,10 @@ def __init__(self, properties):
                     values = np.array(hist['values'])
                 else:
                     # Load the values from a file
-                    values = np.genfromtxt(hist['values']['filename'], delimiter=',')
+                    filename = hist['values']['filename']
+                    if not isabs(filename):
+                        filename = join(directory, filename)
```
bryanwweber: I think this will fail if the input is a `dict` and therefore `directory` is `None`

jsantner: Changing the default directory to `''` would fix that. But, if the input is a dict, then the filename must be specified as an absolute path, right? Since there's no yaml file, a path relative to a yaml file wouldn't make sense. So, line 693 won't be run anyway.

jsantner:
I've never used `Path` objects before, that's an interesting idea. How would the `Path` object be sent from `ChemKED` to `DataPoint` where it's used to read the csv file, though? It still must be passed as an argument, right? I can add that functionality in if you think it's better than passing the directory.

Using a dictionary input, I think the filename would have to be specified as an absolute path. If it were a relative path, what would it be relative to? There's no yaml file.

pyked/chemked.py
bryanwweber: I don't think that change will fix the problem on this line, because the `directory` argument gets set to `None` if `yaml_file` is `None`, so this line will join `None` and `filename`, which won't work. Actually, I think that if people provide a dictionary input, the CSV file has to be specified relative to the PWD of the Python process, or as an absolute file name, but we don't need to check for that case.

bryanwweber:
What if we do the path munging in the loop that creates the `DataPoint`s? Something like
```python
for point in self._properties['datapoints']:
    if 'time-histories' in point:
        for th in point['time-histories']:
            try:
                filename = Path(th['values']['filename'])
            except TypeError:
                pass
            else:
                if yaml_file is not None:
                    th['values']['filename'] = (yaml_file.stem/filename).resolve()
                else:
                    th['values']['filename'] = filename.resolve()
     self.datapoints.append(DataPoint(point))
```
(please correct any indentation errors, writing code with proportional fonts is hard...) Then we don't have to pass anything around. This will resolve the path into an absolute path relative to the yaml file (if given) or relative to the CWD of the Python process, which should handle the dictionary and the yaml file cases gracefully.

BTW, one of the reasons I'm pushing back here is that I don't want to change how `DataPoint` is called, if at all possible.

Can you please make sure to add tests for all these code branches? The diff coverage should be 100%, you can see the lines that haven't been run here: https://codecov.io/gh/pr-omethe-us/PyKED/pull/115/diff where the lines that are in the brighter red in the left column weren't executed during testing.

pyked/chemked.py
jsantner: Good catch. I just pushed another commit so that `directory` will never be set to `None`. Now, if `yaml_file` is `None`, then `directory` will be `''`

jsantner:
That's a smart way to do it, I'll add it in and test it.  
I wasn't sure how to make a test for an error message (line 256 in validation.py), that's why the coverage isn't 100%. Can you point me to an example in the tests where an error message is tested?

bryanwweber:
https://github.com/pr-omethe-us/PyKED/blob/master/pyked/tests/test_chemked.py#L59

CITATION.md
```diff
@@ -11,7 +11,7 @@ A BibTeX entry for LaTeX users is
 ```TeX
 @misc{PyKED,
     author = {Kyle E Niemeyer and Bryan W Weber},
-    year = 2017,
+    year = 2018,
```
kyleniemeyer: It is now 2019

kyleniemeyer:
```suggestion
    year = 2020,
```

kyleniemeyer (commented):
this is a test PR review

PR #130, opened 2021-01-07
Docs deploy

alongd:
This PR adds a GitHub Action for auto docs deployment using [github-pages-action](https://github.com/marketplace/actions/github-pages-action). A workflow file was added along with a requirements file, modifications were made to `.travis.yml`.
When reviewing, please carefully examine the changes made to `.travis.yml` to see that they are only docs-related modifications.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -12,9 +12,6 @@ jobs:
     steps:
       - run: sudo apt-get install pandoc
       - uses: actions/checkout@v2
-      - uses: docker://pandoc/core:2.9
-        with:
-          args: "--help"
 
       - name: Setup Python
         uses: actions/setup-python@v2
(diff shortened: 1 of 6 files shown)


alongd:
@bryanwweber, if you have a minute, please take a look at the error I get during deployment.
Any ideas how to resolve it? Am I missing a package?

image

bryanwweber:
Looks like you need to install Pandoc onto the virtual machine. You should be able to use apt in the Ubuntu images on GitHub actions to do that.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,49 @@
...
+  push:
+    branches:
+      - master
+      - docs_deploy
+
+jobs:
+  deploy:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
```
bryanwweber, line 13: I think you can do something like this:

```suggestion
      - run: sudo apt-get install pandoc
      - uses: actions/checkout@v2
```

Issue #128, opened 2019-10-23
Python 3.7 and 3.8 compatibility 


ndharms:
I'm mainly posting this issue for discussion and to highlight possible todos. 

I was trying to install PyKED and [PyTeCK](https://github.com/pr-omethe-us/PyTeCK) in an environment that requires python 3.7+. I ended up spending a fair amount of time banging my head against the wall with Anaconda until I realized that PyKED (and PyTeCK subsequently) uses python 3.6. Given that Python 3.8 has been released, should there be an effort to convert/modify aspects of PyKED (and PyTeCK) to support 3.7 and 3.8? If so what changes need to be made?

bryanwweber:
I think the only thing that needs to change is that new packages need to be built. Basically, if you bump the build number in the `conda.recipe/meta.yaml` that might work. It's also possible that something like limiting, e.g., Pandas is causing the reliance on 3.6, because there aren't built packages for that old of a dependency on 3.7 and 3.8.

You could probably install with Pip, provided that you installed all the dependencies of PyKED first and then just used pip to install PyKED. Or you could install from source.

kyleniemeyer:
Yeah, I was about to say the same. I imagine/hope most dependencies support 3.7, but 3.8 is probably less likely at this point.

kyleniemeyer:
Quick look at dependencies:
- Pandas supports 3.7
- Pyyaml 3.13 supports 3.7
- numpy v1.17.2 supports 3.7
- I *think* Pint v0.8.1 has a noarch package
- habanero v0.6.0 has noarch
- uncertanties v3.0.3 supports 3.7

The problem might be Cerberus. Right now we only support up versions below 1.2, and v1.1 only supports up to Python 3.6. (It looks like for v1.2 and v1.3.1 now have noarch packages...).

@bryanwweber do you recall why we limit Pint to <0.9?

bryanwweber:
The upper limit on all the dependencies is basically best practice, to avoid breaking changes assuming they follow semver. Since Pint is still [zero-ver](https://0ver.org), it seemed best to limit to the next point release.

kyleniemeyer:
@nateharms if you have the bandwidth, it'd be great if you could check out PyKED's compatibility with newer versions of those dependencies—that will make it easier for us to support Python 3.7+.

But, if cerberus is really limiting us here, then I guess we'll need to revisit the issues limiting us to v1.1

nateharms:
[...]

PR #133, opened 2026-06-09
Refactor

### LINKED ISSUE #131: 0.4.1 (from PyPI) is broken: No such file or directory:  ... /README.md
It attempts to open a non-existent file:
```
===>  Configuring for py38-pyked-0.4.1
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "setup.py", line 11, in <module>
    with open(path.join(here, 'README.md')) as readme_file:
  File "/usr/local/lib/python3.8/codecs.py", line 905, in open
    file = builtins.open(filename, mode, buffering)
FileNotFoundError: [Errno 2] No such file or directory: '/disk-samsung/freebsd-ports/science/py-pyked/work-py38/pyked-0.4.1/README.md'
*** Error code 1
```

kyleniemeyer:
- [x] Fixes #131, #128, #120, (possibly others)
- [x] Tests updated
- [x] Documentation updated / corrected

- big update to modernize PyKED and make it work again / on current Python versions
- prep for v0.5.0

@pr-omethe-us/chemked


bryanwweber:
Heads up that httpx is going through a community breakup right now. There's a fork from Pydantic I'd suggest instead.

kyleniemeyer:
@bryanwweber woof, wonderful. Thanks for the heads up

kyleniemeyer:
@bryanwweber unfortunately it seems `habanero` still depends on `httpx`, so there seems to be not a *ton* of point in switching yet, until they move as well - but I'll note this and perhaps open an issue over there to suggest similar.

kyleniemeyer:
The test suite passes on my machine, but there seem to be DOI/crossref issues when GitHub Actions runs them

kyleniemeyer:
@bryanwweber ah, I see this has been proposed already: https://github.com/sckott/habanero/issues/219

kyleniemeyer:
OK... introducing a fixture that mocks the CrossRef API (via `habanero`) for tests that don't really need to run the DOI helps, but it's still getting rate limited.

Ultimately, we do not really want the tests to fail because of something outside the code, in this case the Crossref API rate-limiting which is leading to a habanero failure. I think the better solution is to mock all external dependencies in the regular test suite, and then add a separate small "live API" test suite that runs on some schedule to make sure nothing external breaks.

kyleniemeyer:
@bryanwweber I think this is good to go now - would welcome any feedback if you have time!

For context, I wanted to get the package working/up to date before @LekiaAnonim works on adding some additional schemas (especially if we redo the approach to schema definition).

src/pyked/orcid.py
```diff
@@ -0,0 +1,28 @@
...
+    Returns:
+        `dict`: Dictionary with the JSON response from the API
+
+    Raises:
+        `~httpx.HTTPStatusError`: If the given ORCID cannot be found, an
+            `~httpx.HTTPStatusError` is raised with status code 404
+    """
+    url = f"https://pub.orcid.org/v2.1/{orcid}/person"
+    r = httpx.get(url, headers=headers)
+    r.raise_for_status()
```
LekiaAnonim, line 27: Compared to the formal code, it seems we are raising status even when the status code is not 200. Was this deliberate?

src/pyked/orcid.py
```diff
@@ -0,0 +1,28 @@
...
+        orcid (`str`): The ORCID to be searched
+
+    Returns:
+        `dict`: Dictionary with the JSON response from the API
+
+    Raises:
+        `~httpx.HTTPStatusError`: If the given ORCID cannot be found, an
+            `~httpx.HTTPStatusError` is raised with status code 404
+    """
+    url = f"https://pub.orcid.org/v2.1/{orcid}/person"
```
LekiaAnonim: There's a newer version of the ORCID api published in 2019. We could use that since we are also upgrading.
https://info.orcid.org/api-version-policy/

> API 3.x (Released May 2019) added new sections to the record such as research resources, simplified the schema structure and provided ORCID with additional flexibility.  It moved much of the validation from the schema into our servers, which has enabled us to fix issues and add new identifier types without releasing further versions of the API.

src/pyked/validation.py
```diff
@@ -177,89 +182,100 @@ def _validate_isvalid_t_range(self, isvalid_t_range, field, values):
...
+        quantity = 1.0 * units(value["units"])
         try:
             quantity.to(property_units[field])
         except pint.DimensionalityError:
-            self._error(field, 'incompatible units; should be consistent '
-                        'with ' + property_units[field]
-                        )
+            self._error(
+                field,
+                "incompatible units; should be consistent " "with " + property_units[field],
```
LekiaAnonim: Could we simplify this using an f-string for readability?

```suggestion
        f"incompatible units; should be consistent with {property_units[field]}",
```

src/pyked/validation.py
```diff
@@ -270,21 +286,22 @@ def _validate_isvalid_quantity(self, isvalid_quantity, field, value):
...
+                    field,
+                    f"value must be greater than 0.0 {property_units[field]}",
                 )
         except pint.DimensionalityError:
-            self._error(field, 'incompatible units; should be consistent '
-                        'with ' + property_units[field]
-                        )
+            self._error(
+                field,
+                "incompatible units; should be consistent " "with " + property_units[field],
```
LekiaAnonim: Could we simplify this using an f-string for readability?

```suggestion
        f"incompatible units; should be consistent with {property_units[field]}",
```

src/pyked/validation.py
```diff
@@ -323,99 +339,109 @@ def _validate_isvalid_reference(self, isvalid_reference, field, value):
...
+                    None,
+                )
                 # error if missing author in given reference information
                 if author_match is None:
-                    self._error(field, 'Missing author: ' +
-                                ' '.join([author['given'], author['family']])
-                                )
+                    self._error(
+                        field,
+                        "Missing author: " + " ".join([author["given"], author["family"]]),
```
LekiaAnonim: Could we simplify this using an f-string for readability?

```suggestion
            f"Missing author: {author['given']} {author['family']}",
```

src/pyked/orcid.py
kyleniemeyer: Good catch, thank you

bryanwweber:
> @bryanwweber I think this is good to go now - would welcome any feedback if you have time!

I'll try to take a look this weekend! Reviewing on my phone is a no-go 😬

src/pyked/orcid.py
kyleniemeyer: Good question - `raise_for_status()` only raises for 4xx/5xx; 2xx responses pass through silently. It's the `httpx` equivalent of the `requests` pattern the original code used

.github/workflows/publish.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Download build artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: dist
+          path: dist/
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          attestations: true
```
bryanwweber, line 49: Is this configured to use the secure token publishing workflow?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,8 +4,8 @@
 - [ ] Documentation updated
 
 Changes proposed in this pull request:
-- 
-- 
-- 
+-
+-
+-
```
bryanwweber: I think these spaces were here intentionally

src/pyked/chemked.py
```diff
@@ -1,30 +1,31 @@
-"""
-Main ChemKED module
-"""
+"""Main ChemKED module"""
+
 # Standard libraries
-from os.path import exists
+import xml.dom.minidom as minidom
+import xml.etree.ElementTree as etree
```
bryanwweber, line 5: We should look at replacing these with, I think, `lxml`. There are CVEs associated with reading untrusted input with the built-in XML library. Not here though.

src/pyked/chemked.py
```diff
@@ -1,30 +1,31 @@
...
-VolumeHistory.time.__doc__ = '(`~numpy.ndarray`): the time during the experiment'
-VolumeHistory.volume.__doc__ = '(`~numpy.ndarray`): the volume during the experiment'
-
-TimeHistory = namedtuple('TimeHistory', ['time', 'quantity', 'type'])
-TimeHistory.__doc__ = 'Time history of the quantity in an RCM experiment'
-TimeHistory.time.__doc__ = '(`~numpy.ndarray`): the time during the experiment'
-TimeHistory.quantity.__doc__ = '(`~numpy.ndarray`): the quantity of interest during the experiment'
+from .validation import Q_, OurValidator, schema, yaml
+
+VolumeHistory = namedtuple("VolumeHistory", ["time", "volume"])
```
bryanwweber: Not necessary, but, there's a newer syntax for `NamedTuple`s that I think would allow the docs more naturally:

```python
from typing import NamedTuple
from numpy.typing import NDArray
...
class VolumeHistory(NamedTuple):
    """Time history..."
    time: NDArray
    """..."""
```

Another option would be a `dataclass` or perhaps even better a pydantic model which IMO would be a better fit for this but weren't widely supported when we started this 😬

src/pyked/chemked.py
```diff
@@ -345,47 +386,54 @@ def convert_to_ReSpecTh(self, filename):
...
+            exp.text = "Ignition delay measurement"
         else:
-            raise NotImplementedError('Only ignition delay type supported for conversion.')
+            raise NotImplementedError("Only ignition delay type supported for conversion.")
 
-        reference = etree.SubElement(root, 'bibliographyLink')
-        citation = ''
+        reference = etree.SubElement(root, "bibliographyLink")
+        citation = ""
         for author in self.reference.authors:
```
bryanwweber: I think this can be simplified:

```suggestion
        citation = ", ".join(self.reference.authors)
```

src/pyked/chemked.py
```diff
@@ -345,47 +386,54 @@ def convert_to_ReSpecTh(self, filename):
...
+            self.reference.journal
+            + " ("
+            + str(self.reference.year)
+            + ") "
+            + str(self.reference.volume)
+            + ":"
+            + self.reference.pages
+            + ". "
+            + self.reference.detail
+        )
```
bryanwweber: This would probably be simpler as an `f` string (and could even be combined with the previous suggestion):

```suggestion
        citation = f"{citation}, {self.reference.journal} ({self.reference.year}) {self.reference.volume}:{self.reference.pages}. {self.reference.detail}"
```

src/pyked/chemked.py
```diff
@@ -394,198 +442,212 @@ def convert_to_ReSpecTh(self, filename):
         # Composition type *has* to be the same
         composition_type = self.datapoints[0].composition_type
         if not all(dp.composition_type == composition_type for dp in self.datapoints):
-            raise NotImplementedError('Error: ReSpecTh does not support varying composition '
-                                      'type among datapoints.'
-                                      )
+            raise NotImplementedError(
+                "Error: ReSpecTh does not support varying composition " "type among datapoints."
```
bryanwweber:
```suggestion
                "Error: ReSpecTh does not support varying composition type among datapoints."
```

src/pyked/chemked.py
```diff
@@ -394,198 +442,212 @@ def convert_to_ReSpecTh(self, filename):
...
+        ]
         time_histories = [getattr(dp, p) for dp in self.datapoints for p in history_types]
-        time_histories = list(filter(None.__ne__, time_histories))
+        time_histories = [th for th in time_histories if th is not None]
 
         if len(self.datapoints) > 1 and len(time_histories) > 1:
-            raise NotImplementedError('Error: ReSpecTh files do not support multiple datapoints '
-                                      'with a time history.')
+            raise NotImplementedError(
+                "Error: ReSpecTh files do not support multiple datapoints " "with a time history."
```
bryanwweber:
```suggestion
                "Error: ReSpecTh files do not support multiple datapoints with a time history."
```

src/pyked/chemked.py
```diff
@@ -844,9 +942,10 @@ def get_cantera_mole_fraction(self, species_conversion=None):
             >>> dp.get_cantera_mole_fraction(species_conversion)
             'h2:4.4400e-03, o2:5.5600e-03, Ar:9.9000e-01'
         """
-        if self.composition_type == 'mass fraction':
-            raise ValueError('Cannot get mole fractions from the given composition.\n'
-                             '{}'.format(self.composition))
+        if self.composition_type == "mass fraction":
+            raise ValueError(
+                "Cannot get mole fractions from the given composition.\n" f"{self.composition}"
```
bryanwweber:
```suggestion
                f"Cannot get mole fractions from the given composition.\n{self.composition}"
```

src/pyked/chemked.py
```diff
@@ -881,9 +980,9 @@ def get_cantera_mass_fraction(self, species_conversion=None):
             >>> dp.get_cantera_mass_fraction(species_conversion)
             'h2:2.2525e-04, o2:4.4775e-03, Ar:9.9530e-01'
         """
-        if self.composition_type in ['mole fraction', 'mole percent']:
-            raise ValueError('Cannot get mass fractions from the given composition.\n'
-                             '{}'.format(self.composition)
-                             )
+        if self.composition_type in ["mole fraction", "mole percent"]:
+            raise ValueError(
+                "Cannot get mass fractions from the given composition.\n" f"{self.composition}"
```
bryanwweber:
```suggestion
                f"Cannot get mass fractions from the given composition.\n{self.composition}"
```

src/pyked/converters.py
```diff
@@ -0,0 +1,719 @@
...
+    # Default version is 0 for the ChemKED file
+    properties["file-version"] = 0
+
+    # Default ChemKED version
+    properties["chemked-version"] = __version__
+
+    return properties
+
+
+def get_reference(root):
```
bryanwweber, line 90: One of the next things we should do if we keep working on this is to type hint the API

src/pyked/converters.py
```diff
@@ -0,0 +1,719 @@
...
+        help='Output filename (e.g., "file1.yaml")',
+    )
+
+    args = parser.parse_args(argv)
+
+    c = chemked.ChemKED(yaml_file=args.input)
+    c.convert_to_ReSpecTh(args.output)
+
+
+def main(argv=None):
```
bryanwweber, line 645: Another improvement would be to consider using a library like `typer` and `rich` to improve the CLI experience

src/pyked/converters.py
```diff
@@ -0,0 +1,719 @@
...
+                "-o",
+                args.output,
+                "-fa",
+                args.file_author,
+                "-fo",
+                args.file_author_orcid,
+            ]
+        )
+
+    elif os.path.splitext(args.input)[1] == ".yaml" and os.path.splitext(args.output)[1] == ".xml":
```
bryanwweber: More future work, switch this to `pathlib.Path` operations.

src/pyked/validation.py
```diff
@@ -323,99 +339,109 @@ def _validate_isvalid_reference(self, isvalid_reference, field, value):
...
-                                    'not match that in reference. Reference: ' +
-                                    orcid + '. Given: ' + author_match['ORCID']
-                                    )
+                                    field,
+                                    author_match["name"]
+                                    + " ORCID does "
+                                    + "not match that in reference. Reference: "
+                                    + orcid
+                                    + ". Given: "
+                                    + author_match["ORCID"],
```
bryanwweber: All these string concatenations would probably be simpler as f-strings.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - id: end-of-file-fixer
+      - id: check-yaml
+        exclude: "^src/pyked/schemas/"
+      - id: check-added-large-files
+
+  - repo: https://github.com/psf/black
+    rev: 26.5.1
+    hooks:
+      - id: black
+
```
bryanwweber: If we have ruff, we don't need black

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    rev: v0.15.16
+    hooks:
+      - id: ruff
+        args: [--fix, --unsafe-fixes]
+
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v2.1.0
+    hooks:
+      - id: mypy
+        files: ^src/
```
bryanwweber: mypy is unlikely to be very useful until we type hint things.

CHANGELOG.md
```diff
@@ -4,18 +4,54 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## [Unreleased]
+## [0.5.0] - 2026-06-09
```
bryanwweber: This should be bumped in a separate PR where we also update the version in `_version.py` and tag that commit.

pyproject.toml
```diff
@@ -0,0 +1,155 @@
...
+]
+docs = [
+    "sphinx",
+    "nbsphinx",
+    "ipython",
+]
+lint = [
+    "pre-commit",
+]
+
```
bryanwweber, line 54: This `lint`, `docs`, and `test` things here should more properly be in a `[dependency-groups]` table: https://pip.pypa.io/en/latest/user_guide/#dependency-groups since they're not intended for end-user support. I'd also suggest moving `pandas` into the main dependencies, it's not hard to install that anymore.

We should also add `ruff` to the `lint` group so it can be run outside of `pre-commit`.

pyproject.toml
```diff
@@ -0,0 +1,155 @@
...
+    "live_api: tests that call external APIs; run with 'pytest -m live_api'",
+]
+addopts = "-m 'not live_api'"
+filterwarnings = [
+    "ignore::pint.UnitStrippedWarning",
+]
+
+[tool.black]
+line-length = 100
+target-version = ["py310", "py311", "py312", "py313", "py314"]
```
bryanwweber: Don't need ruff and black

pyproject.toml
```diff
@@ -0,0 +1,155 @@
...
+
+# https://github.com/charliermarsh/ruff
+[tool.ruff]
+line-length = 100
+target-version = "py314"
```
bryanwweber: I think this should be the lowest version we support

pyproject.toml
```diff
@@ -0,0 +1,155 @@
...
+    "D107", # Missing docstring in __init__
+    "D203", # 1 blank line required before class docstring
+    "D205", # 1 blank line required between summary line and description
+    "D212", # Multi-line docstring summary should start at the first line
+    "D213", # Multi-line docstring summary should start at the second line
+    "D400", # First line should end with a period
+    "D401", # First line should be in imperative mood
+    "D413", # Missing blank line after last section
+    "D415", # First line should end with period, question mark, or exclamation point
+    "D416", # Section name should end with a colon
```
bryanwweber: We're disabling so many docstring ones, why bother adding it in the first place?

pyproject.toml
```diff
@@ -0,0 +1,155 @@
...
+[tool.ruff]
+line-length = 100
+target-version = "py314"
+
+[tool.ruff.lint]
+extend-select = [
+    "E",      # style errors
+    "F",      # flakes
+    "D",      # pydocstyle
+    "I001",   # isort
```
bryanwweber: Suggest just making this `I` for all the isort rules. Also, adding `NPY` for numpy rules, and possibly the pandas (`PD`) ones too.

bryanwweber (approved):
Bunch of small comments, many of them are future work. I'll approve so you can merge when you want to

sckott:
i was waiting on vcrpy to support httpx2 - which was just dropped - so should be able to get a new habanero cut in the next day or two

pyproject.toml
kyleniemeyer: @bryanwweber do you mean for the package overall, Python 3.14?

.github/workflows/publish.yml
kyleniemeyer: @bryanwweber sure is! (This is new)

.github/PULL_REQUEST_TEMPLATE.md
kyleniemeyer: @bryanwweber ah, good point. I think that came from running pre-commit on all files (without protecting this file). I'll revert

kyleniemeyer:
I can see that `habanero` v2.9.1 will support `httpx2`, so will update when this is available on PyPI. Thanks @sckott!

sckott:
@kyleniemeyer up on pypi now

pyproject.toml
bryanwweber: Sorry my comment got cut off 😑 I meant the target version should be the lowest Python version the library supports, 3.10 I think. Otherwise, we risk ruff formatting something that isn't supported by that older version

pyproject.toml
kyleniemeyer: good point, this was left over from an unrelated example. removed all these!

pyproject.toml
kyleniemeyer: ah, got it! fixed this.

pyproject.toml
kyleniemeyer: good call, simplified this and removed black

pyproject.toml
kyleniemeyer: good call, done in f8d3c7b43e851fe5c4cf768798caebc10f235b02

src/pyked/chemked.py
kyleniemeyer: opened this in https://github.com/pr-omethe-us/PyKED/issues/134

src/pyked/chemked.py
kyleniemeyer: Good call - updated the syntax for now in 6c5f7ebe59d0ee5084bc5ac495da807085891e75, but we'll have to consider more-extreme updates separately 😄

src/pyked/converters.py
kyleniemeyer: done in 392e644d0c93d2a8f9af39024dc1ec4ceeebffa1

kyleniemeyer:
OK @bryanwweber, I've taken most of your suggestions and made changes - a few bigger ones are opened as issues. I think this is good to go!

bryanwweber:
:shipit: @kyleniemeyer !

PR #141, opened 2026-07-04
Added new speciation and laminar burning velocity measurement schema

LekiaAnonim:
- [ ] Fixes #
- [x] Tests added/updated
- [x] Documentation updated

- Add ChemKED schema support for laminar burning velocity measurement and speciation measurement datapoints.
- Expand direct-measurement schema fields, apparatus types, and validation support for ignition delay, laminar burning velocity, and speciation measurements.
- Update tests and YAML fixtures to match the current schema, including `equivalence-ratio` value-unit list formatting and direct time-history validation.

Validation performed:

```bash
.venv/bin/python -m pytest -q --tb=short

@pr-omethe-us/chemked

### CODE REVIEW

File: docs/schema-docs.rst [hunk 1]
```diff
@@ -202,23 +301,33 @@ for the :ref:`datapoints <meta-datapoints>` schema.
...
-            * ``temperature``
-            * ``pressure``
-            * ``OH``
-            * ``OH*``
-            * ``CH``
-            * ``CH*``
+        Describes the target measurement (species or physical quantity) used to define ignition.
+        Must be one of: ``temperature``, ``pressure``, ``OH``, ``OH*``, ``CH``, ``CH*``, ``NH3``,
+        ``CO2``, ``N2O``, ``CH4``, ``OHEX``, ``CHEX``, ``CO``, ``H2O``, ``C2``, ``O``,
+        ``CH3OH``, ``CH3``, ``O2``, ``soot``, ``CO;O``, ``[O]*[CO]``, or ``NEOC5H11``.
```
kyleniemeyer: what is the difference between `OH*` and `OHEX`? Similarly, `CH*` and `CHEX`? These seem to be the same despite being named differently, and we don't want to inherit an error/unnecessary differentiation from ReSpecTh. I'm also wondering about `CO;O` and `'[O]*[CO]`, are these the same thing?

LekiaAnonim: OHEX and CHEX could be the same  as CH*,  and OH* respectively. I will remove them including the "CO;O" and "[O]*[CO]"

docs/schema-docs.rst
```diff
@@ -129,18 +140,107 @@ particular experiment type.
...
+* ``temperature``: sequence, optional
+    The temperature of the experiment, with dimensions of temperature. Must conform to
+    :ref:`value-unit-optional <schema-value-unit-optional>`
+
 .. _common-ignition-type:
 
 * ``ignition-type``: mapping, optional
     Has the same schema as :ref:`ignition-type <ignition-ignition-type>`
 
+.. _common-ignition-delay:
```
kyleniemeyer, line 180: is there a situation where multiple data points would be provided with a common ignition delay time? this would normally be the quantity being measured and thus different for each data point... I can't really imagine a scenario where this would be a common value.

docs/schema-docs.rst
```diff
@@ -129,18 +140,107 @@ particular experiment type.
...
+    The ignition delay measurement, with dimensions of time. Must conform to
+    :ref:`value-unit-optional <schema-value-unit-optional>`
+
+.. _common-equivalence-ratio:
+
+* ``equivalence-ratio``: sequence, optional
+    The equivalence ratio of the experiment, dimensionless. Must conform to
+    :ref:`value-unit-optional <schema-value-unit-optional>`
+
+.. _common-laminar-burning-velocity:
```
kyleniemeyer, line 192: same comment as with the common ignition delay time - when would this be a common property?

docs/schema-docs.rst
```diff
@@ -129,18 +140,107 @@ particular experiment type.
...
+
 .. _common-composition:
 
 * ``composition``: mapping, optional
     This mapping provides the specification of the initial composition of the mixture. Fields:
 
     - ``kind``: string, required
-        The ``kind`` can be ``mole fraction``, ``mass fraction``, or ``mole percent``
+        The ``kind`` can be ``mole fraction``, ``mass fraction``, ``mole percent``,
+        ``mol/cm3``, ``mol/m3``, ``mol/L``, or ``mol/dm3``. Fraction kinds must be
```
kyleniemeyer, line 266: `mol/cm3`, `mol/m3`, `mol/L`, and `mol/dm3` are all the same "kind" - molar concentration. They only differ in the *units* of volume used, but I think they should all be the same kind: `molar concentration`.

kyleniemeyer: hmm... now I'm seeing why having these given as different types, since we don't provide a way to indicate units. Not ideal, but I guess this is fine considering how we specify composition.

docs/schema-docs.rst
```diff
@@ -202,23 +308,33 @@ for the :ref:`datapoints <meta-datapoints>` schema.
...
-            * ``temperature``
-            * ``pressure``
-            * ``OH``
-            * ``OH*``
-            * ``CH``
-            * ``CH*``
+        Describes the target measurement (species or physical quantity) used to define ignition.
+        Must be one of: ``temperature``, ``pressure``, ``OH``, ``OH*``, ``CH``, ``CH*``, ``NH3``,
+        ``CO2``, ``N2O``, ``CH4``, ``CO``, ``H2O``, ``C2``, ``O``,
+        ``CH3OH``, ``CH3``, ``O2``, ``soot``.
```
kyleniemeyer, line 343: just noting that we will likely need PyTeCK to flag `soot` as the target, since most kinetic models don't represent soot directly.

docs/schema-docs.rst
```diff
@@ -202,23 +308,33 @@ for the :ref:`datapoints <meta-datapoints>` schema.
...
+            * ``d/dt min extrapolated``: minimum slope of the ``target`` extrapolated to the
+              baseline
+            * ``d/dt max extrapolated``: maximum slope of the ``target`` extrapolated to the
+              baseline
+            * ``d/dt second max``: second maximum of the time derivative of the ``target``
             * ``max``: maximum of the ``target``
             * ``1/2 max``: half-maximum of the ``target``
             * ``min``: minimum of the ``target``
-            * ``d/dt max extrapolated``: maximum slope of the target extrapolated to the baseline
+            * ``concentration``: the ``target`` reaches a specified concentration
```
kyleniemeyer, line 357: should this be more general, like `amount` or `value`? For example, one could in theory specify ignition based on a particular value of temperature or pressure. (Not sure if that's ever used in practice, though, versus `relative increase`, so perhaps fine to leave this way.)

src/pyked/schemas/chemked_schema.yaml
```diff
@@ -38,6 +57,15 @@ apparatus:
       allowed:
         - shock tube
         - rapid compression machine
+        - stirred reactor
+        - jet stirred reactor
+        - tubular reactor
+        - flow reactor # Should we still keep flow reactor when we have tubular reactor and jet stirred reactor?
+        - burner stabilized flame
+        - counterflow flame # ReSpecTh CTF (counterflow twin flame)
```
kyleniemeyer: will we only need to represent counterflow twin premixed flames? "counterflow flame" is not synonymous with "counterflow twin flame", which means a symmetrical premixed configuration. Unless there is a place later to clarify this?

src/pyked/schemas/chemked_schema.yaml
```diff
@@ -38,6 +57,15 @@ apparatus:
       allowed:
         - shock tube
         - rapid compression machine
+        - stirred reactor
```
kyleniemeyer: how is this different than JSR?

src/pyked/schemas/chemked_schema.yaml
```diff
@@ -38,6 +57,15 @@ apparatus:
...
+        - jet stirred reactor
+        - tubular reactor
+        - flow reactor # Should we still keep flow reactor when we have tubular reactor and jet stirred reactor?
```
kyleniemeyer: JSR and flow reactor are different things, but do we need to specify "tubular reactor" as a unique apparatus type or leave it with "flow reactor", which is more general?

src/pyked/schemas/value_unit_schema.yaml
```diff
@@ -8,51 +8,126 @@ value-with-uncertainty: &value-with-uncertainty
...
 value-without-uncertainty: &value-without-uncertainty
   isvalid_quantity: true
   items:
     - anyof_type:
       - string
       - float
+# Metadata-only: just uncertainty/ESD info without a value.
+# Used in common-properties when uncertainty metadata is shared
+# but the property value varies per datapoint.
+value-metadata-only: &value-metadata-only
```
kyleniemeyer: @LekiaAnonim I thought we decided that these should *not* be permitted as a common property and need to be given for each datapoint?

src/pyked/chemked.py
```diff
@@ -184,6 +204,33 @@ def __init__(self, yaml_file=None, dict_input=None, *, skip_validation=False):
         ]:
             setattr(self, prop.replace("-", "_"), self._properties[prop])
 
+    @staticmethod
+    def _apply_common_properties(point, common):
```
kyleniemeyer: again, as per my previous comment, I thought we decided against this

kyleniemeyer (changes requested):
@LekiaAnonim ok, I've gone through and left some comments for you to address.

In addition, I'd like to add complete test files for the new types added, similar to how there are complete shock tube and RCM test files/examples. (And, tests added for validating these files, with the internet-requiring parts mocked.)

src/pyked/schemas/chemked_schema.yaml
kyleniemeyer: actually, I'm going to change my mind a little bit here - "tubular reactor" is probably better than "flow reactor", I guess they are essentially synonymous, so we need to pick one. But I don't have a strong feeling either way

kyleniemeyer:
@rwest @bryanwweber would you both be able to take a look at this extension of the schema?

bryanwweber:
I won't have time, sorry @kyleniemeyer

kyleniemeyer:
@bryanwweber no problem!

@rwest once we have some complete examples are the new measurements and experiment types, we should ask some of our colleagues who specifically work with the relevant experimental apparatuses. They don't need to review the source code here but it would be great to get feedback on the schema / examples.

src/pyked/schemas/chemked_schema.yaml
LekiaAnonim: You're right. Though I was considering JSR as a type of flow reactor, based on nomenclatures in some of the literatures I've reviewed. But we could be specific on the type of reactor to model instead of using reactor classes such as stirred and flow reactors.

src/pyked/schemas/value_unit_schema.yaml
LekiaAnonim: The value-metadata-only attribute will be removed in my next commit. Thanks !

src/pyked/schemas/chemked_schema.yaml
LekiaAnonim: There could exist other apparatus that are not jet-stirred, but which can also be modeled as a perfectly stirred reactor. But it would be best to be specific on the reactor type, that is, capturing exact apparatus types, and not a generic category.

src/pyked/schemas/chemked_schema.yaml
LekiaAnonim: I will change that to counterflow twin premixed flame. There is also a modeling instances on Cantera for counterflow flames with fuel+oxidizer, and opposing product stream.
https://cantera.org/stable/examples/python/onedim/premixed_counterflow_flame.html

docs/schema-docs.rst
LekiaAnonim: One thought will be to convert the concentration into the same unit, then we can use one unit to represent molar concentration across the dataset.

docs/schema-docs.rst
LekiaAnonim: I will remove it from the docs. It's an oversight from an earlier change.

docs/schema-docs.rst
LekiaAnonim: I will remove it from the docs. It's an oversight from an earlier change.

src/pyked/schemas/chemked_schema.yaml
kyleniemeyer: yes, I think this should be specific

src/pyked/schemas/chemked_schema.yaml
kyleniemeyer: As above, these should be listed specifically for what we want/need to support - we can always add additional specific types later.

kyleniemeyer:
@LekiaAnonim just to confirm, I believe this is just waiting on:
- removal of the generic apparatus types
- CHANGELOG entries

LekiaAnonim:
> @LekiaAnonim just to confirm, I believe this is just waiting on:
> 
> * removal of the generic apparatus types
> * CHANGELOG entries

Yes, that's correct.

kyleniemeyer:
Thanks @LekiaAnonim, I think this is ready to merge!

